PR #3, opened 2020-06-22
docs: improve API Docs styling

FRSgit:
add node_modules to gitignore
add sass compiling step to docs & watch makefile tasks
refine docs page styling both on mobile & desktop

references issue https://trello.com/c/f0bRBbxx


FRSgit:
@rlisowski please check if my changes to `package.json` & `makefile` are ok for you. Also if you feel that the scss files should be somewhere else in the project structure, please feel free to point that out.

I've removed `main.css` from the repository. It's not being created via `npm run docs` and the source copy is contained within `scss` files

PR #14, opened 2020-08-11
Show operations headers in Safari

rlisowski:
https://trello.com/c/QhCNIuu4/778-docs-no-operations-headers-in-safari

CSS load order in Safari is different that's why it was invisible only there.

### DIFF
--- a/doc/rapidoc.scss
+++ b/doc/rapidoc.scss
@@ -61,7 +61,7 @@
 .nav-bar-tag {
   margin-left: $spacing5;
   color: $colorPrimaryForeground;
-  display: none;
+  display: flex;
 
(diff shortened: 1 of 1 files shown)


doc/rapidoc.scss
```diff
@@ -61,7 +61,7 @@
 .nav-bar-tag {
   margin-left: $spacing5;
   color: $colorPrimaryForeground;
-  display: none;
+  display: flex;
```
FRSgit, line 64: That's weird. Any idea why it have worked in all other browsers, but not Safari?

rlisowski: css load order in safari is different

PR #28, opened 2020-10-30
Remove all Travis CI configurations

theSoenke:
Remove all Travis CI files


docstun (approved):
@theSoenke thanks! I did this by adding it to `.openapi-generator-ignore` but your approach might be better :)

theSoenke:
@docstun Ah, missed that :) Only saw that the config files still exists in phrase-ruby

PR #35, opened 2020-11-27
Fix CLI readme links

theSoenke:
Fix documentation links from old help center


orwigeg (approved):
Don't tell Kathryn we still had a PhraseApp in there 🙈

PR #37, opened 2020-11-30
Ruby-Client: Support nested objects/hashes in form data

docstun:
Nested objects in POST requests seem broken since objects are not properly serialized when generating the request body. This causes wrong requests, e.g. when submitting `format_options` since the hash is only sent as a string.

Before:

    --------------------------b8e580af624e881d
    Content-Disposition: form-data; name="format_options"
    
    {:ignore_source_translations=>true}
    --------------------------b8e580af624e881d--

After:

    --------------------------7f6d8be7b2cfe386
    Content-Disposition: form-data; name="format_options[ignore_source_translations]"
    
    true
    --------------------------7f6d8be7b2cfe386--

It's not an ideal solution but fixes the issue. For the future we can think about moving all format options to first-class form params of the upload endpoint in the OpenAPI specification.

### DIFF
--- a/openapi-generator/templates/ruby-client/api_client_faraday_partial.mustache
+++ b/openapi-generator/templates/ruby-client/api_client_faraday_partial.mustache
@@ -116,6 +116,8 @@
           when ::Array, nil
             # let Faraday handle Array and nil parameters
             data[key] = value
+          when ::Hash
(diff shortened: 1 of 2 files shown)


theSoenke (approved):
@docstun Thanks for fixing it!

PR #58, opened 2020-12-08
auto-release phrase-cli

egze:


orwigeg (approved):
I think it looks okay - I also tagged @sbruhns  as a reviewer because I think we were talking about something related to one of the changes earlier

PR #60, opened 2020-12-09
autorelease brew

egze:
The plan:

1. Clone `homebrew-brewed` to a tmp folder.
2. Replace file hashes and version in `phrase.rb`.
3. push to master

### DIFF
--- a/clients/cli/build/update_brew.sh
+++ b/clients/cli/build/update_brew.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+set -eo pipefail
+
+brew_dir=$(mktemp -d -t ci-XXXXXXXXXX)
+git clone --depth 1 https://$API_TOKEN_GITHUB@github.com/phrase/homebrew-brewed.git $brew_dir &> /dev/null
+
+# Upload artifacts
+DIST_DIR="./dist"
+FILES_FOR_BREW=("phrase_linux_386.tar.gz" "phrase_linux_amd64.tar.gz" "phrase_macosx_amd64.tar.gz")
+
+# Change the sha256 in Formula/phrase.rb
+for file in ${FILES_FOR_BREW[@]}; do
+  sha256=$(sha256sum "$DIST_DIR/$file" | awk '{ print $1 }')
+  sed -i '' "s/sha256 \".*\" # $file/sha256 \"$sha256\" # $file/g" "$brew_dir/Formula/phrase.rb"
+done
+
+# Change the version in Formula/phrase.rb
+sed -i '' "s/version \".*\"/version \"$VERSION\"/g" "$brew_dir/Formula/phrase.rb"
+
+cd $brew_dir
+git config --global user.email "support@phrase.com"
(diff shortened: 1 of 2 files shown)


clients/cli/build/update_brew.sh
```diff
@@ -0,0 +1,29 @@
...
+# Change the sha256 in Formula/phrase.rb
+for file in ${FILES_FOR_BREW[@]}; do
+  sha256=$(sha256sum "$DIST_DIR/$file" | awk '{ print $1 }')
+  sed -i '' "s/sha256 \".*\" # $file/sha256 \"$sha256\" # $file/g" "$brew_dir/Formula/phrase.rb"
+done
+
+# Change the version in Formula/phrase.rb
+sed -i '' "s/version \".*\"/version \"$VERSION\"/g" "$brew_dir/Formula/phrase.rb"
+
+current=$(pwd)
```
tobstarr, line 20: you could also use `pushd ${pwd}` and then `popd` when done

clients/cli/build/update_brew.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+set -eo pipefail
+
+brew_dir=$(mktemp -d -t ci-XXXXXXXXXX)
+git clone --depth 1 https://$API_TOKEN_GITHUB@github.com/phrase/homebrew-brewed.git $brew_dir &> /dev/null
+
+DIST_DIR="./dist"
+FILES_FOR_BREW=("phrase_linux_386.tar.gz" "phrase_linux_amd64.tar.gz" "phrase_macosx_amd64.tar.gz")
```
tobstarr, line 9: I would have probably just used 
`for file in phrase_linux_386.tar.gz phrase_linux_amd64.tar.gz phrase_macosx_amd64.tar.gz; do`
or maybe even `ls ${DIST_DIR}/phrase_*.tar.gz`? Only a matter of time when we also need `phrase_macosx_arm.tar.gz`.

clients/cli/build/update_brew.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+set -eo pipefail
+
+brew_dir=$(mktemp -d -t ci-XXXXXXXXXX)
+git clone --depth 1 https://$API_TOKEN_GITHUB@github.com/phrase/homebrew-brewed.git $brew_dir &> /dev/null
```
tobstarr, line 6: I would probably now swallow all output here as there could be plenty of reasons why the cloning could not work.

Maybe write the output some where tmp and then cat if the command fails?

tobstarr (approved):

clients/cli/build/update_brew.sh
egze: good to know. will use it next time 👍

clients/cli/build/update_brew.sh
egze: Not all files are needed, for example windows is skipped.

tobstarr (approved):

PR #77, opened 2021-01-19
Added language to current user WORKFLOWS-755

Varpusparvi:
https://phraseapp.atlassian.net/browse/WORKFLOWS-755 

https://github.com/phrase/phrase/pull/7497

### DIFF
--- a/schemas/user.yaml
+++ b/schemas/user.yaml
@@ -13,6 +13,8 @@ user:
       type: string
     position:
       type: string
+    language:
+      type: string
     created_at:
       type: string
       format: date-time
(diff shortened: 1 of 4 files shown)


schemas/user.yaml
```diff
@@ -13,6 +13,8 @@ user:
       type: string
     position:
       type: string
+    language:
```
forelabs, line 16: I am not sure if we should use the `user` object here. Isn't it used for other user returns as well? Then I think we need a new `current_user.yaml` file with a new `current_user` object which is returned by the user#get endpoint

forelabs (approved):
seal-of-approval-white-small

PR #78, opened 2021-01-21
Add support for branching in Screenshots API

docstun:
Allow passing "branch" name since we now support branches in the Screenshots API.


theSoenke (approved):
:tada:

PR #79, opened 2021-02-12
list actions has own ids to make client working

sbruhns:
Batch and single resource endpoints are mapped on the same client action. This will make the batch actions unusable.

In the old client we supported:

```
$ phrase keys delete
```
and 

```
$ phrase key delete
```

In the new client only the plural command is currently supported. This cause `phrase keys delete` is now longer unique and working as expected  https://phraseapp.atlassian.net/secure/RapidBoard.jspa?rapidView=2&projectKey=INTEGRATE&modal=detail&selectedIssue=INTEGRATE-647

This PR solves this by introducing `-collection` commands.

### DIFF
--- a/paths/translations/batch_exclude.yaml
+++ b/paths/translations/batch_exclude.yaml
@@ -1,7 +1,7 @@
 ---
 summary: Set exclude from export flag on translations selected by query
 description: Exclude translations matching query from locale export.
-operationId: translations/exclude
+operationId: translations/exclude-collection
 tags:
 - Translations
 parameters:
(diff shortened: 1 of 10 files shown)


paths/translations/batch_exclude.yaml
```diff
@@ -37,7 +37,7 @@ x-code-samples:
       -H 'Content-Type: application/json'
 - lang: CLI v2
   source: |-
-    phrase translations exclude \
+    phrase translations exclude-collextion \
```
theSoenke, line 40: exclude-collextion -> exclude-collection

sbruhns: Thanks

theSoenke (approved):

PR #87, opened 2021-02-22
Ruby: Use String as default return type

theSoenke:
Use `String` as fallback return_type, which makes the `return_type` parameter optional for the `locale_download` method.


theSoenke:
Will require a more generic solution

PR #92, opened 2021-03-03
Add sort_by param to projects#index [WORKFLOWS-1193]

forelabs:
doc part for https://github.com/phrase/phrase/pull/7692

### DIFF
--- a/paths/projects/index.yaml
+++ b/paths/projects/index.yaml
@@ -8,7 +8,7 @@ parameters:
 - "$ref": "../../parameters.yaml#/X-PhraseApp-OTP"
 - "$ref": "../../parameters.yaml#/page"
 - "$ref": "../../parameters.yaml#/per_page"
-- description: Sort projects. Valid options are "name", "updated_at" and "space".
+- description: Sort projects. Valid options are "name_asc", "name_desc", "updated_at_asc", "updated_at_desc", "space_asc" and "space_desc".
   example:
   name: sort_by
   in: query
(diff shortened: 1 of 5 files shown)


paths/projects/index.yaml
```diff
@@ -8,6 +8,12 @@ parameters:
 - "$ref": "../../parameters.yaml#/X-PhraseApp-OTP"
 - "$ref": "../../parameters.yaml#/page"
 - "$ref": "../../parameters.yaml#/per_page"
+- description: Sort projects. Valid options are "name_asc", "name_desc", "updated_at_asc", "updated_at_desc", "space_asc" and "space_desc".
```
paulbaum, line 11: Not sure if this is the right place but I feel like sort by space is not self explanatory.
Does it sort by space ID or space name? or just groups by space?

forelabs: I think this should have been a question for the phrase pr then :eyes: This pr only introduce the supported sortings to the docs. sorting by an ID does not make sense in my eyes and other then that a space has no other information then a name so I would say there is no need to use "space_name" since there is no other option available.

paulbaum: Could also be by space size (how many projects in there) and since this is in the docs I thought this would be the right place to have a more detailed description on how the params work.
The question was not an actual question but rather pointing out that the customer might want clarification. I know it's about the name since I looked in the code :) 

I thought it'd be nice for the customer to have it explained that space is about the spaces name.
Anyways I'll approve and if you think its obvious to the customer you can just merge it :)

paulbaum (approved):
Feel free to merge it if you think the customer knows whats up with the sort_by_space param :)

PR #99, opened 2021-03-29
make sure to only cleanup file if it is actually initialized

tobstarr:
The function `client.LocalesApi.LocaleDownload` does not create files if there is nothing to write (see https://github.com/phrase/phrase-go/blob/660264e0c1a322325cc83bf42da67b851514c762/client.go#L456). This also means that they cannot be cleaned up.

As we did not check the `err` part of `ioutil.ReadAll(file)` we also did not notice that file was still an uninitialized pointer.

Even though this now works (in case of an uninitialized os.File we now just write the empty byte slice) we should think of changing `client.LocalesApi.LocaleDownload` (as this would probably break the API we might need to add a new function) so that we can e.g. provide an `io.Reader` and so the caller can decide if to actually create a file or maybe if a buffer is also good enough.

### DIFF
--- a/clients/cli/cmd/internal/pull.go
+++ b/clients/cli/cmd/internal/pull.go
@@ -174,9 +174,16 @@ func (target *Target) DownloadAndWriteToFile(client *phrase.APIClient, localeFil
 		}
 	}
 
-	data, _ := ioutil.ReadAll(file)
-	file.Close()
-	os.Remove(file.Name())
+	var data []byte
+	if file != nil {
+		data, err = ioutil.ReadAll(file)
+		if err != nil {
+			return err
+		}
+		file.Close()
+		os.Remove(file.Name())
+	}
+
 	err = ioutil.WriteFile(localeFile.Path, data, 0700)
 	return err
 }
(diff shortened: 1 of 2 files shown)


clients/cli/cmd/internal/pull.go
```diff
@@ -174,9 +174,16 @@ func (target *Target) DownloadAndWriteToFile(client *phrase.APIClient, localeFil
 		}
 	}
 
-	data, _ := ioutil.ReadAll(file)
-	file.Close()
-	os.Remove(file.Name())
+	var data []byte
+	if file != nil {
+		data, err = ioutil.ReadAll(file)
```
sbl, line 179: > As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.

If we're on 1.16 already let's adjust it to go with the recommendation.

sbl (approved):
One nitpick. Functionally this is 👍🏾

clients/cli/cmd/internal/pull.go
tobstarr: I thought about it but I did not want to do the go 1.16 upgrade in this PR.

Deploying the lambda function is still a manual process so it would mean we need force every deployer to upgrade to go1.16 or (would be better) migrate the deployment code to use docker.

Too much of a rabbit whole for this time.

sbl: 👍🏾

theSoenke (approved):

PR #110, opened 2021-05-26
Add Notification and Notification Groups index documentation

aleks:
- Adds index action documentation for Notifcations
- Adds index action documentation for Notification Groups

### DIFF
--- a/schemas/user_preview.yaml
+++ b/schemas/user_preview.yaml
@@ -9,6 +9,4 @@ user_preview:
       type: string
     name:
       type: string
-    email:
-      type: string
   example: 
--- a/schemas/user_preview.yaml
+++ b/schemas/user_preview.yaml
@@ -9,4 +9,6 @@ user_preview:
       type: string
     name:
       type: string
+    email:
+      type: string
   example: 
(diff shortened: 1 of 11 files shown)


schemas/user_preview.yaml
```diff
@@ -9,4 +9,6 @@ user_preview:
       type: string
     name:
       type: string
+    email:
```
forelabs, line 12: I think we have removed that property again from the previews? If not, I need to check again where this is added. I was asked that the email is only exposed through the user details, not the preview.

forelabs (changes requested):
Please remove the email expose within the user preview here, we have to clarify that first and keep it seperated.

The fix for the account and account details is fine, thanks for that.

schemas/user_preview.yaml
aleks: @forelabs it was removed from GraphQL i think, but we expose `email` on various endpoints (e.g. user, member, everything that is using `UserSerializer`). I added it back in, since the API is still returning this data.

forelabs (approved):

PR #115, opened 2021-06-04
add name field to orders with example for create

paulbaum:
Adding the name field to orders in the docs
it is not a required field since it always get set before saved(before validations even)

https://github.com/phrase/phrase/pull/8015
https://phraseapp.atlassian.net/browse/WORKFLOWS-1741


forelabs (approved):
LGTM

PR #117, opened 2021-06-11
Account locales [WORKFLOWS-1766]

forelabs:
List account locales

image


paths/account_locales/index.yaml
```diff
@@ -0,0 +1,42 @@
+---
+summary: List locales used in account
+description: List all locales used across all projects within an account.
+operationId: account/locales
```
aleks, line 4: @forelabs i didn't notice the file path here, but i think this should go in ` paths/accounts/locales.yaml`

forelabs: I had this first but was not sure, I thought I found something else which had also a new folder though it was part of a different path. But yeah from the path point of view it makes more sense. I can change that, no hard feelings about that.

paths/accounts/locales.yaml
```diff
@@ -0,0 +1,42 @@
...
+- "$ref": "../../parameters.yaml#/X-PhraseApp-OTP"
+responses:
+  '200':
+    description: OK
+    content:
+      application/json:
+        schema:
+          type: array
+          items:
+            "$ref": "../../schemas/locale_preview.yaml#/locale_preview"
```
aleks, line 17: @forelabs i also noticed, that this type will not match, if the suggestion from this PR will be added: https://github.com/phrase/phrase/pull/8044#pullrequestreview-682653475

aleks (approved):
noice!

PR #120, opened 2021-06-15
Keys: Batch Exclude/Include in locale

forelabs:
Two new endpoints to allow batch excluding and including locales on keys.

### DIFF
--- a/paths/keys/exclude.yaml
+++ b/paths/keys/exclude.yaml
@@ -1,0 +1,78 @@
+---
+summary: Exclude a locale on a collection of keys
+description: Exclude a locale on keys matching query. Same constraints as list.
+operationId: keys/exclude
+tags:
+- Keys
+parameters:
+- "$ref": "../../parameters.yaml#/X-PhraseApp-OTP"
+- "$ref": "../../parameters.yaml#/project_id"
+responses:
+  '200':
+    description: OK
+    content:
+      application/json:
+        schema:
+          "$ref": "../../schemas/affected_resources.yaml#/affected_resources"
+    headers:
+      X-Rate-Limit-Limit:
+        "$ref": "../../headers.yaml#/X-Rate-Limit-Limit"
+      X-Rate-Limit-Remaining:
+        "$ref": "../../headers.yaml#/X-Rate-Limit-Remaining"
+      X-Rate-Limit-Reset:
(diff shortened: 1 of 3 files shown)


paths/keys/exclude.yaml
```diff
@@ -0,0 +1,78 @@
...
+      -u USERNAME_OR_ACCESS_TOKEN \
+      -X PATCH \
+      -d '{"branch":"my-feature-branch","q":"mykey* translated:true","locale_id":"abcd1234abcd1234abcd1234abcd1234","tags":"landing-page,release-1.2"}' \
+      -H 'Content-Type: application/json'
+- lang: CLI v2
+  source: |-
+    phrase keys exclude \
+    --project_id <project_id> \
+    --data '{"branch":"my-feature-branch", "query":"'mykey* translated:true'", "locale_id":"abcd1234abcd1234abcd1234abcd1234", "tags":"landing-page,release-1.2"}' \
+    --access_token <token>
```
aleks, line 43: @forelabs i think the params and descriptions in this PR are both incorrect? Looks like a copy of the search, but AFAIR these endpoints only accept locale/key ids as params?

aleks (changes requested):

paths/keys/exclude.yaml
forelabs: @aleks it is a batch endpoint. It accepts locale and then a query to get a set of keys. loading is exactly the same as e.g. for tag/untagging keys.
see https://developers.phrase.com/api/#patch-/projects/{project_id}/keys/tag

```
q:
Specify a query to do broad search for keys by name (including wildcards).

The following qualifiers are also supported in the search term:

ids:key_id,... for queries on a comma-separated list of ids
name:key_name for text queries on exact key names - whitespaces need to be prefixed with a backspace (\"\\\")
tags:tag_name to filter for keys with certain tags
translated:{true|false} for translation status (also requires locale_id to be specified)
updated_at:{>=|<=}2013-02-21T00:00:00Z for date range queries
unmentioned_in_upload:upload_id to filter keys unmentioned within upload
```

aleks: @forelabs ah, i didn't see, that this was handled by the method which gets the keys. Alright! 👍

aleks (approved):

PR #125, opened 2021-06-25
Add documentation for account-level search

aleks:
Updates documentation to include search across projects.

### DIFF
--- a/paths/accounts/search.yaml
+++ b/paths/accounts/search.yaml
@@ -1,0 +1,76 @@
+---
+summary: Search across projects
+description: Search for keys and translations in all account projects
+operationId: account/search
+tags:
+- Accounts
+parameters:
+- "$ref": "../../parameters.yaml#/X-PhraseApp-OTP"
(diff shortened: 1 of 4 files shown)


paths/accounts/search.yaml
```diff
@@ -0,0 +1,76 @@
+---
+summary: Search across projects
+description: Search for keys and translations in all account projects
+operationId: account/search
+tags:
+- Accounts
```
forelabs, line 6: I think this will put `Search across projects` under `Accounts` in the sidebar?
I am not sure if this is where we should put it. To me it is probably even a new sidebar header "Search".
The result of an endpoint should reflect some resource of the header, so to me everything below `Accounts` should return some sort of account (short version, detail, list, ...) or have mutations on accounts.

forelabs (changes requested):
Looks good. I think we should introduce a new namespace called "Search" where this new endpoint lives in (within the api doc navigation). What do you think?

forelabs (approved):
seal-of-approval-white-small

PR #128, opened 2021-07-07
UploadAPI error : fix by using new Utils::tryFopen and Query::build methods

jeromemacias:
`\GuzzleHttp\Psr7\try_fopen` and `\GuzzleHttp\Psr7\build_query` does not exists anymore on the last GuzzleHttp version


theSoenke (approved):
Thanks!

PR #130, opened 2021-07-08
phrase-php: Fix upload_create object params was not correctly sent

jeromemacias:
On UploadApi and create method, params `format_options` and `locale_mapping` cannot be sent due to a fatal error.

So now, for example, we can pass ['my_option' => 'my-value'] as format_options argument and it will send as :
name `format_options[my_option]` and contents `my-value`.

### DIFF
--- a/openapi-generator/templates/php/api.mustache
+++ b/openapi-generator/templates/php/api.mustache
@@ -536,7 +536,7 @@ use {{invokerPackage}}\ObjectSerializer;
             $formParams['{{baseName}}'] = \GuzzleHttp\Psr7\Utils::tryFopen(ObjectSerializer::toFormValue(${{paramName}}), 'rb');
             {{/isFile}}
             {{^isFile}}
-            $formParams['{{baseName}}'] = ObjectSerializer::toFormValue(${{paramName}});
+            $this->formParamsAppend($formParams, '{{baseName}}', ${{paramName}});
             {{/isFile}}
         }
         {{/formParams}}
@@ -664,6 +664,21 @@ use {{invokerPackage}}\ObjectSerializer;
         }
 
         return $options;
+    }
+    /**
+     * Append to form params, handle object params
+     */
+    protected function formParamsAppend(&$formParams, $name, $value)
+    {
+        if (is_object($value)) {
+            foreach ((array) $value as $k => $v) {
+                $formParams[$name.'['.$k.']'] = ObjectSerializer::toFormValue($v);
+            }
+
+            return;
+        }
+
+        $formParams[$name] = ObjectSerializer::toFormValue($value);
     }
 }
 {{/operations}}


egze:
HI @jeromemacias Thanks for taking your time to contribute. I created a task for our next sprint to check your PR. You will hear something in a week or two.

orwigeg:
Hey @jeromemacias - Would it work for you to try formatting the `format_options` in your request like the example below? And if this still causes an error for you, could you provide us some more information on the fatal error you're getting?

```
$format_options = new stdClass;
$format_options-> enable_pluralization = True;
$format_options = json_encode((array)$format_options);
```

jeromemacias:
@orwigeg Thanks for the help on this issue.
Unfortunalty, I already tried this and just retry and I do not have any error, upload works but my format options are all ignored :(

orwigeg:
Hi @jeromemacias. Thank you for the further details, we are now able to recreate the issue with the `upload_create` and the `format_options`. Unfortunately, the solution you've provided (passing in a list) uses a format for the `format_options` that strays from the one we have documented and have been using, that is standard for us. If you would like to modify the solution to add the params in the form that was previously expected, feel free to do so and we would be happy to merge it, and if not, we can also make a ticket for this in _our_ backlog to create the fix.

This is an example of what would need to work:
```
$format_options = new stdClass;
$format_options-> enable_pluralization = False;
$format_options = json_encode((array)$format_options);
```

jeromemacias:
Hi, I will be happy to change my PR but I think I missed somthing.

In the documentation https://github.com/phrase/phrase-php/blob/HEAD/docs/Api/UploadsApi.md#parameters, `format_options` is of type `Object`, so stdClass in PHP, but if I understand your previous comment, the param should pass as JSON string like this: `$format_options = '{"foo": "bar"}'` ?

orwigeg:
Hey @jeromemacias, that's a good point. Ideally you could just pass in a `stdClass` as the docs are already saying. The existing PHP client docs are not great yet and something we're looking into to improve. But the ideal behavior would be if you could just pass in a `stdClass` into the library with the format options.

jeromemacias:
Ok, so my proposal accept an stdClass and an array : https://github.com/phrase/openapi/pull/130/files#diff-ded5c744ec29bc4d6563f2bac8794d1294b05172d6cb265cbd779372619a6ce0R673

I should remove the array compatibility / detection and this PR could be valid ?

orwigeg:
Hey @jeromemacias - That sounds like a plan, if you remove the array compatibility, we would be happy to accept it 🙂

jeromemacias:
@orwigeg It's done ;)

orwigeg:
Thanks @jeromemacias - we merged your branch and are creating a new release.

jeromemacias:
Thanks @orwigeg !

PR #131, opened 2021-07-12
[GROWTH-895] Add API documentation for Job Templates

francawinter:
Description
Jobs Templates is reflected in API Doc

Ticket
https://phraseapp.atlassian.net/browse/GROWTH-895


francawinter:
I think that some requestBody properties can be deleted, e.g. `due_date`, `ticket_url`, `translation_key_ids` for `paths/job_templates/create.yaml`.

paths/job_template_locales/show.yaml
```diff
@@ -0,0 +1,54 @@
+---
+summary: Get a single job template locale
+description: Get a single job template locale for a given job template.
```
vermaxik, line 3: As I can see it's not possible to get/create/delete/update without project_id in `Job Template Locales` section without `project_id`.

Would we like to add this in description e.g.
```suggestion
description: Get a single job template locale for a given project and job template.
```

what do you think?

vermaxik:
I was able to run docs locally, looks good 👍, but I can't really verify if all params in endpoints are correct. 

And I found also misleading in responses in [code](https://github.com/phrase/phrase/pull/8105/files) and docs.

Regarding the implementation of Template Job endpoint in successful response for:

- show 
- create
- update

should return the same response like for `show` (from `::JobTemplates::DetailSerializer`), but now it's different.

What I would suggest to QA the docs with a real request with all params using this documentation. It would be easiest way to detect if the docs not correct. What do you think?

francawinter:
@vermaxik i finished to review every enpoint, please check punctually if it is fine now :) thanks for the great teamwork. Do you think i should remove the generated files (upload_cleanup.go etc) from the PR or can I commit it?

vermaxik:
> @vermaxik i finished to review every enpoint, please check punctually if it is fine now :) thanks for the great teamwork. Do you think i should remove the generated files (upload_cleanup.go etc) from the PR or can I commit it?

I've merged the master to your branch and recompile json. Generated files were gone, because it's already in master in other change.

vermaxik:
First of all I would like to say that you did an incredible job ❤️ 👍 🚀  thanks a lot for that. 

Most of things  looks perfectly now and consistence 👌 , but something I noticed when comparing it to the phrase code - I'll write to separate comments.

paths/job_template_locales/create.yaml
```diff
@@ -0,0 +1,76 @@
...
+        properties:
+          branch:
+            description: specify the branch to use
+            type: string
+            example: my-feature-branch
+          locale_id:
+            description: locale id
+            type: string
+            example: abcd1234cdef1234abcd1234cdef1234
+          user_ids:
```
vermaxik, line 68: from code I can see also one more params param `reviewer_ids`

paths/job_template_locales/update.yaml
```diff
@@ -0,0 +1,78 @@
...
+        properties:
+          branch:
+            description: specify the branch to use
+            type: string
+            example: my-feature-branch
+          locale_id:
+            description: locale id
+            type: string
+            example: abcd1234cdef1234abcd1234cdef1234
+          user_ids:
```
vermaxik, line 70: from code I can see also one more params param `reviewer_ids`

paths/job_templates/index.yaml
```diff
@@ -0,0 +1,72 @@
...
+  source: |-
+    curl "https://api.phrase.com/v2/projects/:project_id/job_templates?branch=my-feature-branch&owned_by=abcd1234cdef1234abcd1234cdef1234&assigned_to=abcd1234cdef1234abcd1234cdef1234" \
+      -u USERNAME_OR_ACCESS_TOKEN
+- lang: CLI v2
+  source: |-
+    phrase job templates list \
+    --project_id <project_id> \
+    --branch my-feature-branch \
+    --owned_by abcd1234cdef1234abcd1234cdef1234 \
+    --assigned_to abcd1234cdef1234abcd1234cdef1234 \
```
vermaxik, line 65: Can't see in the code filter like `owned_by` and `assigned_to`, I think we need to remove it.

paths/job_templates/destroy.yaml
```diff
@@ -0,0 +1,44 @@
...
+      -d '{"branch":"my-feature-branch"}' \
+      -H 'Content-Type: application/json'
+- lang: CLI v2
+  source: |-
+    phrase job templates delete \
+    --project_id <project_id> \
+    --id <id> \
+    --branch my-feature-branch \
+    --access_token <token>
+- lang: CLI v1
```
vermaxik, line 41: We don't support CLI v1 anymore, please remove it everywhere from those examples 🎉

paths/job_template_locales/update.yaml
francawinter: it is reflected in the user's `role` now 👍

vermaxik (approved):
it's amazing 💛  🌴 🚀  thanks for the contribution!

PR #133, opened 2021-07-15
Improve documentation for key name escaping

theSoenke:
Add all keyname characters that have to be escaped to the docs


intro.md
```diff
@@ -859,7 +859,7 @@ The following qualifiers are also supported in the search term:<br />
 
 <ul>
 <li><code>ids:key_id,...</code> for queries on a comma-separated list of ids</li>
-<li><code>name:key_name</code> for text queries on exact key names - whitespaces need to be prefixed with a backspace ("\")</li>
+<li><code>name:key_name</code> for text queries on exact key names - whitespaces, commas, and colons  need to be escaped with a backslash</li>
```
oskarols-phrase, line 862: Looks like there's double spaces after "colons". Shouldn't it be "double backslashes"?

theSoenke: Thanks. Both should be fixed

oskarols-phrase (approved):
Good stuff! 👍

PR #139, opened 2021-07-23
Fix account/locales operation definition

jeromemacias:
Parameter `account_id` was missing into definition. 
So clients `Locale::accountLocales` methods cannot works.


theSoenke:
Thanks for the PR :raised_hands:  In this case, we should use the `id` parameter as it is not called `account_id` in the path. I've changed this in https://github.com/phrase/openapi/pull/141

PR #143, opened 2021-08-11
Pre select the Tag for the uploaded keys

GMetaxakis:
Suggestion PR, to tag uploaded keys with the phrase push

https://github.com/phrase/phrase-cli/pull/74

### DIFF
--- a/clients/cli/cmd/push.go
+++ b/clients/cli/cmd/push.go
@@ -22,6 +22,7 @@ func initPush() {
 				Wait:               params.GetBool("wait"),
 				Branch:             params.GetString("branch"),
 				UseLocalBranchName: params.GetBool("use-local-branch-name"),
+				Tag:                params.GetString("tag"),
 			}
 			err := cmdPush.Run()
 			if err != nil {
@@ -34,5 +35,6 @@ func initPush() {
 	AddFlag(pushCmd, "bool", "wait", "w", "Wait for files to be processed", false)
 	AddFlag(pushCmd, "string", "branch", "b", "branch", false)
 	AddFlag(pushCmd, "bool", "use-local-branch-name", "", "push from the branch with the name of your currently checked out branch (git or mercurial)", false)
+	AddFlag(pushCmd, "string", "tag", "", "Pre tag the uploaded keys with a specific tag", false)
 	params.BindPFlags(pushCmd.Flags())
 }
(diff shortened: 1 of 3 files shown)


egze:
Hi @GMetaxakis Thanks for taking your time to contribute. I created a task for our next sprint to check your PR. You will hear something in a week or two.

GMetaxakis:
> Hi @GMetaxakis Thanks for taking your time to contribute. I created a task for our next sprint to check your PR. You will hear something in a week or two.

Hey @egze any update?

theSoenke:
Hi @GMetaxakis, just a couple of questions about your use case. Do you want to be able to dynamically add tags to all pushed translation keys (in case some of the keys and translations already exist in Phrase) or only those for which translations have changed during a push? And how are you then later using the tag?

Each upload also receives a tag in the form `upload-<identifier>` which can be helpful in some cases.

GMetaxakis:
In our CI we are creating a job for all the new keys that got created after the `phrase push `. This to happen we are fetching the response from the `uploads ` API and we are checking the new keys number that exist there and the last uploaded tag. This work OK for now for Android as we do only 1 upload (on iOS we do 2 for strings and stringsdict), but we are exploring ways of modularization the phrase pushes to different android modules, again it kind of works great in the config file with the tag option, but we cannot check the last 5 uploads or 50 last uploads (if we go with 50 modules with copies) if we have new keys or what is the tag to create the job. Does it make sense?

So in a way we want only for the new keys a predefined tag to create the job instead of searching in all the uploads what is this tag.

clients/cli/cmd/push.go
```diff
@@ -34,5 +35,6 @@ func initPush() {
 	AddFlag(pushCmd, "bool", "wait", "w", "Wait for files to be processed", false)
 	AddFlag(pushCmd, "string", "branch", "b", "branch", false)
 	AddFlag(pushCmd, "bool", "use-local-branch-name", "", "push from the branch with the name of your currently checked out branch (git or mercurial)", false)
+	AddFlag(pushCmd, "string", "tag", "", "Pre tag the uploaded keys with a specific tag", false)
```
theSoenke, line 38:
```suggestion
	AddFlag(pushCmd, "string", "tag", "", "Tag uploaded keys", false)
```

Would be great if you could adapt the description like this

theSoenke:
@GMetaxakis Thanks for the explanation! That sounds very useful. Left one comment on your PR. Everything else looks good to me

theSoenke (approved):

theSoenke:
Thanks for adding this to the CLI :tada:  This is now part of CLI 2.3.0 https://github.com/phrase/phrase-cli/releases/tag/2.3.0

PR #145, opened 2021-09-03
Fix cache file permissions

theSoenke:
Fix issue when the cached version state can not be written and use 666 for file permissions

https://github.com/phrase/phrase-cli/issues/86

Screenshot from 2021-09-03 16-28-21


vermaxik (approved):
LGTM 👍  thanks for fixing 🚀

PR #146, opened 2021-09-06
INTEGRATE-1429: Add `lock` and `unlock` endpoint definitions to OpenAPI

orwigeg:
[INTEGRATE-1429](https://memsourcegroup.atlassian.net/browse/INTEGRATE-1429)

### DIFF
--- a/paths/jobs/lock_job.yaml
+++ b/paths/jobs/lock_job.yaml
@@ -1,0 +1,40 @@
+---
+summary: Lock a job
+description: If you are the job owner, you may lock a job using this API request.
+operationId: job/keys/lock
+tags:
+- Jobs
+parameters:
+- "$ref": "../../parameters.yaml#/X-PhraseApp-OTP"
+- "$ref": "../../parameters.yaml#/project_id"
+- "$ref": "../../parameters.yaml#/id"
+- description: specify the branch to use
+  example: my-feature-branch
+  name: branch
+  in: query
+  schema:
+    type: string
+responses:
+  '204':
+    "$ref": "../../responses.yaml#/204"
+  '400':
+    "$ref": "../../responses.yaml#/400"
+  '404':
+    "$ref": "../../responses.yaml#/404"
+  '429':
+    "$ref": "../../responses.yaml#/429"
+x-code-samples:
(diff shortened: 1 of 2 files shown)


docstun:
@orwigeg is this still relevant?

orwigeg:
@docstun - It depends. David said we should wait until the job locking and unlocking feature were available outside of a feature toggle. It looks like it's still in the "beta" column in `feature_toggle.rb`. If we don't have any plans on removing it from there, perhaps we can just close the PR.

orwigeg:
@docstun - I just got word from Sönke that we got the greenlight to release job locking while I was away on vacation. We'll move the ticket to our new TSS board and we can keep this PR open and merge it soon.

docstun:
@orwigeg thanks a lot - happy to hear!

paths/jobs/lock_job.yaml
```diff
@@ -0,0 +1,40 @@
+---
+summary: Lock a job
+description: If you are the job owner, you may lock a job using this API request.
+operationId: job/keys/lock
```
theSoenke, line 4: I wonder whether this should be `job/lock`?

paths/jobs/unlock_job.yaml
```diff
@@ -0,0 +1,40 @@
+---
+summary: Unlock a job
+description: If you are the job owner, you may unlock a locked job using this API request.
+operationId: job/keys/unlock
```
theSoenke, line 4: same as above

paths/jobs/lock_job.yaml
orwigeg: Hm, good point

orwigeg: I'll look into this tomorrow morning

orwigeg: I feel like there was a reason we did it this way, but I can't remember what it was

orwigeg: I looked into it again this morning and for the life of me I can't remember what our reasoning was, so I'll change it :)

theSoenke (approved):
:tada:

PR #154, opened 2021-10-20
[Bugfix] Ignored tag of locale

GMetaxakis:
After the merge of the #143 a bug was introduced to the system that if we were using the cli with `tag` param, it was ignoring the `tag` that was passed to the locale file.


theSoenke (approved):
@GMetaxakis Thank you for providing the fix. Looks good :heavy_check_mark:

theSoenke:
The change is now released in v2.4.1

PR #158, opened 2021-12-09
Add `complete_review` endpoint and `reviewer_ids` param to job locales

vermaxik:
Improve documentation:
- added POST `/projects/{project_id}/jobs/{job_id}/locales/{id}/complete_review`
- added `reviewer_ids` param to `Create a job locale` and `Update a job locale`
- recompile all docs with previous commits


paths/job_locales/complete_review.yaml
```diff
@@ -0,0 +1,59 @@
+---
+summary: Review a job locale
+description: Marked Job locale as reviewed.
```
theSoenke, line 3: I'd change it to `Mark job locale as reviewed`

schemas/job_locale.yaml
```diff
@@ -28,9 +28,10 @@ job_locale:
       name: My Job 1
       state: completed
     users:
-    - id: abcd1234cdef1234abcd1234cdef1234
+      id: abcd1234cdef1234abcd1234cdef1234
```
theSoenke, line 31: I think the dash is required as it's a list right?

vermaxik: good point 👍  thanks, returned back

paths/job_locales/complete_review.yaml
vermaxik: 👍  changed, thx

PR #160, opened 2021-12-21
Update the UI library 

vermaxik:
The UI improvement of the docs:

- RapiDoc is updated to 9.1.3
- "Try" button is added
- Bearer Authentication
- A path to endpoints are changed to editor's friendly format (removed curly brackets)


oskarols-phrase:
It looks a little bit like we updated some dependency without committing the change to package.json, could this be the case?

There's e.g. changes to the rapidoc minified file and a big update to `yarn.lock`

vermaxik:
> It looks a little bit like we updated some dependency without committing the change to package.json, could this be the case?
> 
> There's e.g. changes to the rapidoc minified file and a big update to `yarn.lock`

Good catch, thanks, fixed.

oskarols-phrase:
LGTM at a glance, nice job 👍

PR #161, opened 2021-12-21
🔥 Clean up the CLI v1 examples

vermaxik:
Preparation to bunch of API Docs improvements: 
- Phrase CLI v1 is deprecated for years, examples are removed from docs and regenerated.


theSoenke (approved):
Nice! Less confusion with just one CLI example. I'm wondering whether we should also remove the v2 suffix as there is no separation vs v1 and v2 anymore

vermaxik:
> Nice! Less confusion with just one CLI example. I'm wondering whether we should also remove the v2 suffix as there is no separation vs v1 and v2 anymore

Thx, good point, but I would leave it for now with versioning because it's still mentioned on https://phrase.com/cli/ 
This should be changed in a both projects.

PR #169, opened 2022-01-13
Add support for Ruby 3

docstun:
Update link-header-parser dependency and restrict to use with Ruby 2.6+ which is required by link-header-parser.


theSoenke (approved):
Great! Thanks for fixing it

PR #173, opened 2022-01-25
CLI: Add project_id to uploads cleanup command

theSoenke:
Add project_id flag for `uploads cleanup` command in cases in which there is no `.phrase.yml` config or the config does not contain a default project id


forelabs (approved):
seal-of-approval-white-small

PR #179, opened 2022-02-04
CLI: Use correct project id for upload cleanup

theSoenke:
It's using the wrong project id if a manual project id is provided. Missed adding it here: https://github.com/phrase/openapi/pull/173


StefanGreffenius (approved):
🌴

PR #186, opened 2022-03-08
Add compiled json for job template id

orwigeg:


theSoenke (approved):
We should probably add a lint step to make sure the compiled.json is always up to date in the future :grimacing:

PR #199, opened 2022-04-25
Fix uploading screenshots via CLI

theSoenke:
The templates for the CLI did not handle cases correctly in which a parameter starts with `file` like `filename` or `fileencoding`. Now files should be opened and non-file types treated correctly as well. In addition, changing the screenshots content-type to `multipart/form-data`

### DIFF
--- a/doc/compiled.json
+++ b/doc/compiled.json
@@ -738,30 +738,7 @@
           "teams": {
             "type": "array",
             "items": {
-              "type": "object",
-              "title": "team",
-              "properties": {
-                "id": {
-                  "type": "string"
-                },
-                "name": {
-                  "type": "string"
-                },
-                "created_at": {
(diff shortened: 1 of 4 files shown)


sbruhns (approved):

doc/compiled.json
```diff
@@ -738,30 +738,7 @@
...
-                  "format": "date-time"
-                }
-              },
-              "example": {
-                "id": "2e7574e8f2372906a03110c2a7cfe671",
-                "name": "My first Team",
-                "created_at": "2020-02-25T12:17:25Z",
-                "updated_at": "2020-03-13T14:46:57Z"
-              }
+              "$ref": "#/components/schemas/team_short"
```
sbruhns, line 741: Good change but seems not really related

PR #203, opened 2022-05-05
Fix quotes for boolean params in examples

theSoenke:
Remove quotes around boolean values in examples and simplify the project create and update example


forelabs (approved):
LGTM

PR #204, opened 2022-05-05
Terminology changes (excluding resource renaming)

forelabs:
**Ticket**
- https://memsource.tpondemand.com/entity/71011-investigate-api-terminology-change-issues

> The terminology changes will add breaking change to all our clients because methods will be renamed.
For example `PhraseAPI::BlacklistedKeys` will get renamed to `PhraseAPI::BlockedKeys` in *phrase-ruby*
or in *phrase-cli* `phrase blacklisted_keys list` will become `phrase blocked_keys list`.

Because of that this pr will **only change terminology except resource renaming** to avoid breaking changes.

image
image

image

### DIFF
--- a/doc/compiled.json
+++ b/doc/compiled.json
@@ -55,8 +55,7 @@
       "name": "Comments"
     },
     {
-      "name": "Distributions",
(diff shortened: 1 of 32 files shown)


forelabs:
@docstun Since we alredy quickly talked about it:

- I would revert the resource name changes so that those stay the same
- We probably would need to add a note to the description that e.g. Glossary is now Term base in the ui

docstun:
@forelabs we cannot break APIs in a minor version. So if we want to rename the actual commands, we must release it as a new major version.

sbruhns (approved):

PR #211, opened 2022-06-20
Add support for fallback locale to locales#create and locales#update

sbruhns:
https://memsource.tpondemand.com/entity/73263-api-allow-fallback_locale_id-parameter-in-localecreate

### DIFF
--- a/doc/compiled.json
+++ b/doc/compiled.json
@@ -5677,11 +5677,11 @@
         "x-code-samples": [
           {
             "lang": "Curl",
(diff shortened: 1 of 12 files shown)


doc/compiled.json
```diff
@@ -5727,6 +5727,11 @@
                     "type": "string",
                     "example": "abcd1234abcd1234abcd1234abcd1234"
                   },
+                  "fallback_locale_id": {
+                    "description": "Fallback locale. Can be the name or public id of the locale. Preferred is the public id.",
```
theSoenke, line 5731: I think we can change it to just `id` instead of public id. As an API user there is really only one type of id

theSoenke: Maybe let's write something like "Fallback locale for empty translations. Can be a locale name or id"?

theSoenke (approved):

PR #220, opened 2022-08-04
CLI: Fix passing in host as CLI argument

theSoenke:
Use `host` and `access_token` args when calling `phrase init --host <host> --access_token <token>`

### DIFF
--- a/clients/cli/cmd/internal/init.go
+++ b/clients/cli/cmd/internal/init.go
@@ -87,11 +87,16 @@ type InitCommand struct {
 
 func (cmd *InitCommand) Run() error {
 	// keep host if specified in config file or as command line parameter
-	if cmd.Config.Credentials.Host != "" {
-		cmd.YAML.Host = cmd.Config.Credentials.Host
+	if cmd.Config.Host != "" {
+		cmd.YAML.Host = cmd.Config.Host
 	}
 
-	step := StepAskForToken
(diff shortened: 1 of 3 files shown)


clients/cli/cmd/internal/init.go
```diff
@@ -87,11 +87,16 @@ type InitCommand struct {
 
 func (cmd *InitCommand) Run() error {
 	// keep host if specified in config file or as command line parameter
-	if cmd.Config.Credentials.Host != "" {
-		cmd.YAML.Host = cmd.Config.Credentials.Host
+	if cmd.Config.Host != "" {
```
strzibny, line 90: Can I ask the difference between:

- `cmd.Config.Credentials`
- `cmd.Config`, and
- `cmd.Credentials`

It confuses me a little.

theSoenke: This also took me a while to understand all these different fields at first. What happens here is that the `Config` struct is inheriting the fields from `Credentials` struct:

```golang
type Config struct {
  Credentials
}
```

Now you can access the `Credentials` fields via `config.Credentials.Host` or directly `config.Host`, which contain the same value

strzibny: I see, thx.

strzibny (approved):

PR #222, opened 2022-08-05
Allow empty translation content to be send

strzibny:
Currently it's not possible to send an empty translation which might be important if we want to use a machine translation later.

This removes the `omitempty` option from the `content` field by providing a custom JSON options via `x-json-tag`.

I changed the template to the intent. So the tag now provides a replacement.

This is how a generated model looks like:

```
package phrase

import (
	"time"
)

// Translation struct for Translation
type Translation struct {
	Id           string        `json:"id,omitempty"`
	Content      string        `json:"content"`
	Unverified   *bool         `json:"unverified,omitempty"`
	Excluded     *bool         `json:"excluded,omitempty"`
	PluralSuffix string        `json:"plural_suffix,omitempty"`
	Key          KeyPreview    `json:"key,omitempty"`
	Locale       LocalePreview `json:"locale,omitempty"`
	Placeholders []string      `json:"placeholders,omitempty"`
	State        string        `json:"state,omitempty"`
	CreatedAt    time.Time     `json:"created_at,omitempty"`
	UpdatedAt    time.Time     `json:"updated_at,omitempty"`
}
```

### DIFF
--- a/openapi-generator/templates/go/model.mustache
+++ b/openapi-generator/templates/go/model.mustache
@@ -34,7 +34,7 @@ type {{classname}} struct {
 {{#description}}
 	// {{{description}}}
 {{/description}}
-	{{name}} {{#isNullable}}*{{/isNullable}}{{^isNullable}}{{#isBoolean}}*{{/isBoolean}}{{/isNullable}}{{{dataType}}} `json:"{{baseName}}{{^required}},omitempty{{/required}}"{{#withXml}} xml:"{{baseName}}{{#isXmlAttribute}},attr{{/isXmlAttribute}}"{{/withXml}}{{#vendorExtensions.x-go-custom-tag}} {{{.}}}{{/vendorExtensions.x-go-custom-tag}}`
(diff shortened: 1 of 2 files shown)


openapi-generator/templates/go/model.mustache
```diff
@@ -34,7 +34,7 @@ type {{classname}} struct {
 {{#description}}
 	// {{{description}}}
 {{/description}}
-	{{name}} {{#isNullable}}*{{/isNullable}}{{^isNullable}}{{#isBoolean}}*{{/isBoolean}}{{/isNullable}}{{{dataType}}} `json:"{{baseName}}{{^required}},omitempty{{/required}}"{{#withXml}} xml:"{{baseName}}{{#isXmlAttribute}},attr{{/isXmlAttribute}}"{{/withXml}}{{#vendorExtensions.x-go-custom-tag}} {{{.}}}{{/vendorExtensions.x-go-custom-tag}}`
+	{{name}} {{#isNullable}}*{{/isNullable}}{{^isNullable}}{{#isBoolean}}*{{/isBoolean}}{{/isNullable}}{{{dataType}}} `{{#vendorExtensions.x-go-custom-tag}}{{{.}}}{{/vendorExtensions.x-go-custom-tag}}{{^vendorExtensions.x-go-custom-tag}}json:"{{baseName}}{{^required}},omitempty{{/required}}"{{#withXml}} xml:"{{baseName}}{{#isXmlAttribute}},attr{{/isXmlAttribute}}"{{/withXml}}{{/vendorExtensions.x-go-custom-tag}}`
```
theSoenke, line 37: These templates are a bit difficult to read but it removes `omitempty` when a the custom tag is set right? If not sure how extensible the schema is but should we maybe introduce a new field like `go-allow-empty` to the schema and set `omitempty` based on that boolean value? Otherwise using `x-go-custom-tag` will now always drop `omitempty` which might not be wanted in every case

strzibny: Yes, I realize this downside. I can try to rename it and see if just making a new one works.

strzibny: Okay, different names worked too, so then I could redo it much more cleanly with `x-json-tag` which replaces only the JSON part.

The generated model is the same.

theSoenke (approved):

PR #228, opened 2022-09-27
Rebrand to Phrase Strings

strzibny:


theSoenke (approved):
We also need to release new version of the CLI and libraries but let's merge this change for now

Issue #225, opened 2022-08-30
PHP 8.1 support


smoench:
Currently you are using openapi-generator in version 4.8.1. This generator produces a PHP client which is only compatible with PHP < 8.1. The openapi-generator [already fixed it with version 6.x](https://github.com/OpenAPITools/openapi-generator/pull/11743). You probably want to update your php templates as in here https://github.com/OpenAPITools/openapi-generator/tree/master/modules/openapi-generator/src/main/resources/php. Once updated it will require PHP 7.4 and later.

PR #234, opened 2022-09-29
Fix PHP 8.1 deprecations

smoench:
solves #225 

Fixes `Deprecated: Return type of Phrase\\Model\\Locale::offset*($offset) should either be compatible with ArrayAccess::offset*(mixed $offset): void, or the #[\\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice` deprecation with PHP 8.1.


smoench:
@theSoenke Can you maybe look into this here? Thanks in advance :-)

forelabs:
@theSoenke a new php release will be required

theSoenke:
Thanks for the fix @smoench! Just released it with version 1.5.2

smoench:
Thank you @theSoenke!

PR #238, opened 2022-10-13
Switch Branding

orwigeg:
Remove the parrot and add the new Phrase logo


theSoenke (approved):

forelabs:
please link the ticket on your prs :slightly_smiling_face: 

https://memsource.tpondemand.com/entity/79540-rebranding-replace-phrase-logo-at-phrase

theSoenke:
@forelabs It's a public repo and I think on public ones we should not necessarily post our internal links, which are only helpful for us internally

forelabs:
> @forelabs It's a public repo and I think on public ones we should not necessarily post our internal links, which are only helpful for us internally

Hmm makes sense but also Internal URLs are secured. So i think in this case please attach the PR to the ticket manually.

PR #249, opened 2022-11-10
Introduce --batch option for CLI

jablan:
https://memsource.tpondemand.com/entity/74406-orch-support-option-in-phrase-cli

### DIFF
--- a/clients/cli/cmd/internal/push.go
+++ b/clients/cli/cmd/internal/push.go
@@ -14,6 +14,7 @@ import (
 	"github.com/phrase/phrase-cli/cmd/internal/paths"
 	"github.com/phrase/phrase-cli/cmd/internal/placeholders"
 	"github.com/phrase/phrase-cli/cmd/internal/print"
+	"github.com/phrase/phrase-cli/cmd/internal/shared"
 	"github.com/phrase/phrase-cli/cmd/internal/spinner"
 	"github.com/phrase/phrase-go/v2"
 )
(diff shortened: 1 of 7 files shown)


egze:
Can't really comment much on the implementation. Maybe let's get it live and test it in action?

jablan:
@egze could you use just the binary from the branch, as sort of a QA run?

clients/cli/cmd/internal/push.go
```diff
@@ -190,8 +194,7 @@ func (source *Source) Push(client *phrase.APIClient, waitForResults bool, branch
 				print.Failure("There was an error processing %s. Your changes were not saved online.", localeFile.RelPath())
 			}
 		} else {
-			fmt.Println("done!")
-			fmt.Printf("Check upload Id: %s, filename: %s for information about processing results.\n", upload.Id, upload.Filename)
+			print.NonBatchPrintf("done!\nCheck upload Id: %s, filename: %s for information about processing results.\n", upload.Id, upload.Filename)
```
forelabs, line 197: I think we should have a json response here as well.
How about printing the whole `upload` object?

clients/cli/cmd/internal/upload_cleanup.go
```diff
@@ -79,7 +85,7 @@ func UploadCleanup(client *phrase.APIClient, cmd *UploadCleanupCommand) error {
 			return err
 		}
 
-		fmt.Printf("%d key(s) successfully deleted.\n", affected.RecordsAffected)
+		print.Success(fmt.Sprintf("%d key(s) successfully deleted.\n", affected.RecordsAffected))
```
forelabs, line 88: I think we should return json here, maybe the whole `affected` object?

forelabs (approved):

PR #252, opened 2022-11-21
Add x-cli-version attribute to all operations

jablan:
https://memsource.tpondemand.com/entity/80698-orch-support-openapi-spec-provides-cli_version

I opted for succinct value, without the patch number (as patch updates don't introduce API changes) and without the leading `>` (as any other symbol makes no sense).

### DIFF
--- a/paths/accounts/index.yaml
+++ b/paths/accounts/index.yaml
@@ -41,4 +41,4 @@ x-code-samples:
   source: |-
     phrase accounts list \
     --access_token <token>
-
+x-cli-version: '2.5'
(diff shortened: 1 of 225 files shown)


paths/accounts/index.yaml
```diff
@@ -41,4 +41,4 @@ x-code-samples:
   source: |-
     phrase accounts list \
     --access_token <token>
-
+x-cli-version: '2.5'
```
forelabs, line 44: has it been discussed with the orch team that we change to simply the version number without a prefix symbol?

forelabs (approved):

paths/accounts/index.yaml
egze: It depends what does this field mean. If it means that `2.5` is the minimal CLI version that support the endpoint fully, then it's OK.

jablan: yes, that's what it would mean. I think, until we start deprecating some endpoints, that should cover the needs.

PR #253, opened 2022-11-30
Document API limitations

theSoenke:
Add documentation for API limitations


strzibny (approved):
I am not 100% on smaller bold text. I think a regular cursive would look better, but not a blocker.

Screenshot from 2022-12-05 14-28-12

theSoenke:
@strzibny Great suggestion and should be updated now with italic formatting

PR #255, opened 2022-12-07
Create tar.gz archive for Linux arm64 as well

der-eismann:
Currently the Linux arm64 release is only available as a binary and not as an archive, this will save some transfer bytes.


theSoenke (approved):
Thanks for the PR. Looks good :heavy_check_mark:

theSoenke:
It's now part of the latest release https://github.com/phrase/phrase-cli/releases/tag/2.6.1

PR #258, opened 2023-01-06
Add media and point-of-contact to project API

jablan:
https://memsource.tpondemand.com/entity/84624-api-add-main-technology-and-point


paths/projects/create.yaml
```diff
@@ -56,6 +56,10 @@ requestBody:
             description: Main file format specified by its API Extension name. Used for locale downloads if no format is specified. For API Extension names of available file formats see <a href="https://support.phrase.com/hc/en-us/sections/6111343326364">Format Guide</a> or our <a href="#formats">Formats API Endpoint</a>.
             type: string
             example: yml
+          media:
```
theSoenke, line 59: Should we maybe include the allowed values here?

paths/projects/update.yaml
```diff
@@ -58,10 +58,18 @@ requestBody:
             description: "(Optional) Name of the project"
             type: string
             example: My Android Project
+          point_of_contact:
+            description: (Optional) User ID of the point of contact for the project. Pass `nil` to unset.
```
theSoenke, line 62: nit: but maybe we should replace `nil` with `null` as it's json

paths/projects/create.yaml
theSoenke: Ah, just saw the dropdown and assumed those are the only allowed

theSoenke (approved):

theSoenke (approved):

PR #259, opened 2023-01-11
CLI: pull should separate tags by files

jablan:
Addresses https://github.com/phrase/phrase-cli/issues/105

Also eliminates multiple identical API calls when multiple tags are requested


theSoenke (approved):
LGTM. Just tested it and works really well!

PR #263, opened 2023-01-16
Fix linux/arm64 Docker build

theSoenke:
Pushing the image twice is overwriting the previous image. Instead both platforms needs to used in a single build/push command

### DIFF
--- a/clients/cli/build/release.sh
+++ b/clients/cli/build/release.sh
@@ -18,8 +18,12 @@ IMAGE=${IMAGE_PREFIX}:${VERSION}
 IMAGE_LATEST=${IMAGE_PREFIX}:latest
 
 echo build docker image "${IMAGE}" and ${IMAGE_LATEST}
-docker buildx build --tag "${IMAGE}" --tag ${IMAGE_LATEST} --build-arg CLI_PATH="dist/phrase_linux_amd64" --platform linux/amd64 -f ./Dockerfile --push .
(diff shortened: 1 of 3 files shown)


clients/cli/build/release.sh
```diff
@@ -18,8 +18,12 @@ IMAGE=${IMAGE_PREFIX}:${VERSION}
 IMAGE_LATEST=${IMAGE_PREFIX}:latest
 
 echo build docker image "${IMAGE}" and ${IMAGE_LATEST}
-docker buildx build --tag "${IMAGE}" --tag ${IMAGE_LATEST} --build-arg CLI_PATH="dist/phrase_linux_amd64" --platform linux/amd64 -f ./Dockerfile --push .
-docker buildx build --tag "${IMAGE}" --tag ${IMAGE_LATEST} --build-arg CLI_PATH="dist/phrase_linux_arm64" --platform linux/arm64 -f ./Dockerfile --push .
+
+mkdir -p dist/linux
+cp dist/phrase_linux_amd64 dist/linux/amd64
+cp dist/phrase_linux_amd64 dist/linux/arm64
```
jablan, line 24: typo in first argument, should be `arm64`

jablan (changes requested):

clients/cli/build/release.sh
theSoenke: Thanks! This could have gone wrong. Without an M1 Mac I was not really able to test it

jablan (approved):

PR #268, opened 2023-01-18
Remove old brew package updates

theSoenke:
Disable brew auto updates for the deprecated tap and only ignore auto-generated CLI files


jablan (approved):
thanks for the `.gitignore` change

PR #269, opened 2023-01-19
Add deployment diagram

jablan:


theSoenke (approved):
phrase-ruby and phrase-js are still missing but otherwise this looks really good

jablan:
good point, added these too

theSoenke (approved):

PR #271, opened 2023-01-23
Allow pushing files without extensions

jablan:
https://memsource.tpondemand.com/entity/84434-phrase-cli-error-because-of-messages

Not sure why this check was here in the first place. Unfortunately, git blame is not very informative and there are no specs covering this validation.
Currently I see no drawbacks in removing this check.


theSoenke:
@jablan I tried to push `simple_json` file with a file extension, which then results in a 422. Maybe we should also consider whether the server side validation still is useful or not

Screenshot from 2023-01-23 17-13-32

theSoenke (approved):

PR #274, opened 2023-02-02
Update to OpenAPI generator v2

theSoenke:
- Upgrade to the generator package v2 (but keep the generator cli at 4.3.1 to avoid breaking changes)
- Add missing resources to schema and fix linting warnings
- Update licenses


schemas.yaml
```diff
@@ -110,6 +110,14 @@ schemas:
     "$ref": schemas/job_preview.yaml#/job_preview
   job_locale:
     "$ref": schemas/job_locale.yaml#/job_locale
+  job_template:
```
jablan, line 113: how come we were missing all these?

openapi-generator/go_lang.yaml
```diff
@@ -2,7 +2,7 @@
 generatorName: go
 outputDir: clients/go
 packageName: phrase
-packageVersion: 2.6.5
+packageVersion: 2.6.6
```
jablan, line 5: is the version bump just due to tooling upgrade? should there be any actual change in the generated library?

jablan (approved):

openapi-generator/go_lang.yaml
jablan: ah I see the comment now.

schemas.yaml
theSoenke: honestly i don't know either. running `npx openapi-generator-cli validate -i tmp/compiled.yaml` will find all of these issues but it also finds a few more deprecated things i've not been able to fix so far

PR #277, opened 2023-02-06
Document team feature for job and job template locales

orwigeg:


paths/job_locales/create.yaml
```diff
@@ -73,4 +73,18 @@ requestBody:
...
+            example:
+            - abcd1234cdef1234abcd1234cdef1234
+          reviewer_team_ids:
+            description: Array of team ids to be assigned to the job locale as reviewers
+            type: array
+            items:
+              type: string
+            example:
+            - abcd1234cdef1234abcd1234cdef1234
 x-cli-version: '2.5'
```
theSoenke, line 90: I think this needs to be set to an CLI release that will include the new params. So maybe 2.6.7 if that's the next release

theSoenke (commented):
I think the `compiled.json` also would need to be regenerated and library version bumped

paths/job_locales/create.yaml
```diff
@@ -73,4 +73,18 @@ requestBody:
               type: string
             example:
             - abcd1234cdef1234abcd1234cdef1234
+          team_ids:
```
strzibny, line 76: This should be called `translator_team_ids`

strzibny:
Can we include changes for the `#show` action too?

paths/job_locales/create.yaml
orwigeg: In the ticket on TP, it specified the parameter should be called `team_ids`

orwigeg:
> Can we include changes for the `#show` action too?

Yes, this was still a WIP, but that was the missing piece :)

paths/job_locales/create.yaml
strzibny: I believe it will be teams for the `#show` where teams combine translators and reviewers

strzibny: This ticket has the good specification, where do you see it? @orwigeg 

Screenshot from 2023-02-16 09-11-50

strzibny:
@theSoenke I tried to bring this up to date with what we actually have.

paths/job_locales/create.yaml
```diff
@@ -73,4 +73,20 @@ requestBody:
               type: string
             example:
             - abcd1234cdef1234abcd1234cdef1234
-x-cli-version: '2.5'
+          translation_team_ids:
```
theSoenke, line 76: I think this should be `translator_team_ids` instead of `translation_team_ids` right?

strzibny: Yes, of course! Will fix.

jablan:
Could we use `refs` to avoid duplication?

in `paths/job_locales/update.yaml`:

```yaml
          translator_team_ids:
            $ref: "create.yaml#/requestBody/content/application~1json/schema/properties/translator_team_ids"
```

strzibny:
@jablan Like this? https://github.com/phrase/openapi/pull/277/commits/25430a95652966058d9cc314a3ce9f8cc1a8d3a1

jablan:
Yep, IMO it reduces duplication and we should prefer it. @theSoenke what do you think?

strzibny:
@theSoenke do you think it's good to go at this point?

paths/job_template_locales/create.yaml
```diff
@@ -73,6 +73,20 @@ requestBody:
...
+            description: Array of team ids to be assigned to the job locale as reviewers
+            type: array
+            items:
+              type: string
+            example:
+            - abcd1234cdef1234abcd1234cdef1234
         required:
         - locale_id
-x-cli-version: '2.5'
+x-cli-version: '2.6.7'
```
theSoenke, line 92: Should be 2.6.8 now as 2.6.7 has been released this morning :sweat_smile:

theSoenke:
@strzibny @jablan Using refs seems good but there were some issues with them in the past. When I run the lint step now I can also see some new issues:

```
make bundle
npx openapi-generator-cli validate -i tmp/compiled.yaml
```

```
[main] WARN  o.o.codegen.utils.ModelUtils - Failed to get the schema name: #/paths/~1projects~1%7Bproject_id%7D~1job_templates~1%7Bjob_template_id%7D~1locales/post/requestBody/content/application~1json/schema/properties/reviewer_team_ids
```

paths/job_template_locales/create.yaml
strzibny: Will update :)

strzibny:
I decided to revert my ref change for now & updated the version.

paths/job_template_locales/update.yaml
```diff
@@ -75,6 +75,18 @@ requestBody:
...
+            example:
+            - abcd1234cdef1234abcd1234cdef1234
+          reviewer_team_ids:
+            description: Array of team ids to be assigned to the job locale as reviewers
+            type: array
+            items:
+              type: string
+            example:
+            - abcd1234cdef1234abcd1234cdef1234
+x-cli-version: '2.6.7'
```
theSoenke, line 92: @strzibny sorry, but there's one more place with the old version :see_no_evil:

theSoenke (approved):

PR #284, opened 2023-02-21
Revert x/net update

jablan:


openapi-generator/cli_lang.yaml
```diff
@@ -2,4 +2,4 @@
 generatorName: go
 outputDir: clients/cli
 packageName: phrase
-packageVersion: 2.6.7
+packageVersion: 2.6.8
```
theSoenke, line 5: This should probably stay at 2.6.6 right? As that's the latest release

jablan: somehow I thought we should never go backwards when declaring releases. especially now as there has been a commit in between.

theSoenke: i think that should be fine. the release step should check if there is already a tag for the version and if that's the case it will not try to create a new release as far as I know

jablan: ok, but 2.6.7 then, as we are supposed to publish it?

theSoenke: I thought if the change is reverted there is no change we really need to publish and therefore still keeping the old version would make sense

theSoenke (approved):

PR #292, opened 2023-02-24
Improve linting and check for compile errors

theSoenke:
- Fail on validation errors from `npx openapi-generator-cli validate`
- Check if the go and cli code is compiling
- Check `compiled.json` is up to date

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,5 +1,5 @@
 name: Lint
-on: [push, pull_request]
+on: [push]
 jobs:
   lint:
     name: Lint
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -12,6 +12,17 @@ jobs:
       - run: npm install
       - run: npm run lint
 
+  compare-output:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v1
+        with:
+          node-version: "18"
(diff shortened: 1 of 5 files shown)


.github/workflows/lint.yml
```diff
@@ -1,13 +1,64 @@
...
+        with:
+          go-version: 1.19.6
+      - name: Build
+        env:
+          API_TOKEN_GITHUB: ${{ secrets.API_TOKEN_GITHUB }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          npm install
+          npm run generate.go
+          cd clients/go && go install ./...
```
jablan, line 44: any reason for using `install` over `build`?

jablan (approved):

.github/workflows/lint.yml
theSoenke: `go build` just works as well here and changed it. `go install` is also copying the output to `$GOPATH/bin` and maybe it doesn't even do anything when there is no main function like in  this library

Makefile
```diff
@@ -11,8 +11,8 @@ run:
 	@echo Client libraries generated, check clients directory.
 lint:
 	npx swagger-cli validate main.yaml
-	# make bundle
```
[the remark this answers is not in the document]
theSoenke: yes, i didn't wanted to make the build output confusing by printing a lot of warnings we're ignoring for now anyway

PR #298, opened 2023-02-27
fix: disable compile check

theSoenke:
Disabling the compile check for the CLI for now. This step fails when for example a new endpoint is being added as we first need to release a new version of phrase-go for the cli to compile


bikmazefe (approved):
🙌

jablan:
:chicken: :egg:

PR #299, opened 2023-02-27
Update go lib dependencies

jablan:


theSoenke (commented):
@jablan Maybe it's also good to bump the phrase-go version? We also need a new version to bump the library in the cli

docstun:
@jablan any update on this?

PR #305, opened 2023-04-14
Fix the CLI example for job comments

vermaxik:
We use cli v2 examples to generate docker comments for phrase cli.

Fix to use correct one.


theSoenke (approved):
Thanks for fixing! :heavy_check_mark:

PR #311, opened 2023-05-04
[Java] remove multiple README templates

jablan:


theSoenke (approved):
Can't really tell whether any of these are still important but if the code generation still works it should probably be all good

jablan:
It's quite weird. There are other files that seem duplicated too (like Java classes). Agree, probably some artifact of auto-generation of the templates.

PR #320, opened 2023-05-31
feat: Add release-please for CHANGELOG management #TSI-1713

jablan:
This PR introduces [release-please](https://github.com/google-github-actions/release-please-action) for CHANGELOG and version management of the client libraries.

Whenever a PR has been merged to master, a set of github actions are launched, and they open a PR for every client which has been affected by the changes in the commits since the last version. These PRs contain
* a `CHANGELOG.md` in corresponding `clients/...` directory
* bump of the client version in `openapi-generator/templates/..._lang.yml` configuration files

The "affecting" changes are all the changes in the common part of the repo (for example, updates to the API itself) and all the changes in corresponding `/openapi-generator/templates/...` subdirectory. So, changes only affecting a single client will not end up as PRs in other clients.

For proper version and changelog generation, it is important to use conventional commit naming. Another [github action](https://github.com/marketplace/actions/semantic-pull-request) has been added to ensure that PRs have proper titles, which would end up as commit messages upon squash commits.


openapi-generator/python_lang.yaml
```diff
@@ -3,7 +3,10 @@ generatorName: python
 outputDir: clients/python
 packageName: phrase_api
 projectName: phrase-api
-packageVersion: 1.9.0
```
jablan, line 6: This should be managed by release-please and is moved to `version.txt` in python template directory.

Makefile
```diff
@@ -39,7 +39,7 @@ go:
 typescript:
 	openapi-generator-cli generate -i tmp/compiled.yaml -g typescript-fetch -o clients/typescript -t ./openapi-generator/templates/typescript-fetch -c ./openapi-generator/typescript_lang.yaml
 python:
-	openapi-generator-cli generate -i tmp/compiled.yaml -g python -o clients/python -t ./openapi-generator/templates/python -c ./openapi-generator/python_lang.yaml
```
jablan, line 42: specified in configuration file instead

.github/workflows/build.yml
```diff
@@ -153,6 +153,7 @@ jobs:
           npm run generate.python
           git clone https://$API_TOKEN_GITHUB@github.com/phrase/phrase-python.git clones/python &> /dev/null
           rsync -avI --delete --exclude='.git/' clients/python/ clones/python
+          cp openapi-generator/templates/python/CHANGELOG.md clones/python/
```
jablan, line 156: it's not included when generating Python client so needs to be copied manually

openapi-generator/templates/python/README.mustache
```diff
@@ -5,7 +5,7 @@ Phrase Strings is a translation management platform for software projects. You c
 This Python package is automatically generated by the [OpenAPI Generator](https://openapi-generator.tech) project:
 
 - API version: {{appVersion}}
-- Package version: {{packageVersion}}
```
jablan, line 8: loaded from a partial (from `version.txt`) instead from the configuration file

openapi-generator/templates/python/version.mustache
```diff
@@ -0,0 +1 @@
+version.txt
```
jablan, line 1: symlink

openapi-generator/templates/java/libraries/google-api-client/build.gradle.mustache
```diff
@@ -24,7 +24,7 @@ if(hasProperty('target') && target == 'android') {
 
     apply plugin: 'com.android.library'
     apply plugin: 'com.github.dcendents.android-maven'
-
+
```
theSoenke: nit but should these added whitespaces be here?

theSoenke (commented):
@jablan Looks good to me as far as I can tell! Could you maybe also add a PR description to add how it works and can be used?

And got a couple of questions:
- How does release-please knows which component are affected. For example when we add a new endpoint how would the PR title need to look like?
- Is there also a way to add the changelog to the release in the library repos?
- It creates one PR per library release right? Maybe this could also be automated?

openapi-generator/templates/java/libraries/google-api-client/build.gradle.mustache
jablan: hm that change has been reverted, make sure to check the latest state of the PR

jablan:
@theSoenke I added the description. Answers to your questions:

* PR title affects only whether the version bump is release, minor or major, and what is written to the CHANGELOG. In order to know which component is affected, release-please relies on the directory where the files are changed. As in our case, these changes happen on the very top of the tree, I had to use the root (`.`) in all the actions, and relying on exclusion lists to mark those subdirectories that _should not_ be taken into account when spotting a change. That's why the configuration files are a bit unwieldy and repetitive. Ideally there would be a single configuration file with many components listed, but unfortunatelly I couldn't find a way to achieve this, so there are N configuration files for every client.
* Hm, I haven't checked that. I don't see an easy way, but perhaps we can find one. Right now, the users will have to resort to reading the CHANGELOG.md in corresponding library repo.
* What do you mean by automated? As written above, ideally there would be a single PR with multiple CHANGELOGs altered, but I couldn't find a way to do it, considering the organization of our repo. The config file doesn't support multiple packages with the same root (which is our case).

theSoenke:
@jablan cool thanks. That already helps a lot. Brought up the last question because it seems a bit repetitive if we need to need to approve and merge 7 PRs for API changes. Looks like pushing directly to main does not seem to be supported by release-please right now. But as a workaround maybe we could add another trigger that automerges those PRs?

jablan:
@theSoenke yes, that seems like the only possibility right now. I would actually prefer to have a manual step in between, but ideally that would be only 1 PR with all the changelog updates in it. anyway, I would leave the automatic merge for after this ticket (being that during this QW or not) as I'd like to see how this thing works in practice first for some time.

theSoenke:
@jablan sounds like a plan to at least manually check for a while

theSoenke (approved):

PR #323, opened 2023-06-01
CLI: Fix batch printing for ratelimit message

theSoenke:
Print rate limit warning as JSON when `batch` mode is enabled. Timer is also no updated inline anymore as the color printing was interfering with that

### DIFF
--- a/clients/cli/cmd/internal/print/print.go
+++ b/clients/cli/cmd/internal/print/print.go
@@ -52,11 +52,21 @@ func NonBatchPrintf(msg string, args ...interface{}) {
 
 func Success(msg string, args ...interface{}) {
 	if shared.BatchMode {
-		jsonStuct := &InfoMessage{Message: fmt.Sprintf(msg, args...)}
-		encodedJson, _ := json.Marshal(jsonStuct)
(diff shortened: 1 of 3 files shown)


clients/cli/cmd/internal/print/print.go
```diff
@@ -52,14 +52,24 @@ func NonBatchPrintf(msg string, args ...interface{}) {
...
 }
 
+func Warn(msg string, args ...interface{}) {
+	if shared.BatchMode {
+		jsonStruct := &InfoMessage{Message: fmt.Sprintf(msg, args...)}
+		encodedJson, _ := json.Marshal(jsonStruct)
+		fmt.Println(string(encodedJson))
+	} else {
+		WithColor(ct.Yellow, msg, args...)
+	}
```
jablan, line 70: I guess it makes sense to extract this now into a function like `BatchSensitivePrint(string, color, args)` or something?

jablan (approved):

clients/cli/cmd/internal/print/print.go
theSoenke: Was also thinking whether it should be extract as the building the JSON seems quite similar. But for example the error message looks differently from the other messages. That makes unifying more difficult :thinking:

PR #348, opened 2023-06-20
docs: rename endpoint summaries to be unique

flys1ck:
Actions in Orchestrator are showing up under the same name. Since we do not show any related categories, it was confusing what the action actually does.

image

##### What has been done

* rename endpoint summaries


jablan:
I think this should actually be `fix` rather than `docs` as it reflects in the clients too (`phrase help spaces`). @theSoenke what do you think?

PR #358, opened 2023-07-11
feat(api): Add index API endpoint for custom_metadata_labels

anthonyshchang:
### DONT MERGE
DONT MERGE UNTIL FEATURE TOGGLE IS REMOVED!

### Ticket
https://memsourcegroup.atlassian.net/browse/TSS-1890

### DIFF
--- a/schemas/custom_metadata_label.yaml
+++ b/schemas/custom_metadata_label.yaml
@@ -21,6 +21,10 @@ custom_metadata_label:
         - text
     user:
       "$ref": "./user_preview.yaml#/user_preview"
+    projects:
+      type: array
+      items:
+        "$ref": "./project_short.yaml#/project_short"
     created_at:
       type: string
       format: date-time
@@ -38,3 +42,9 @@ custom_metadata_label:
       name: Joe Doe
     created_at: '2015-01-28T09:52:53Z'
     updated_at: '2015-01-28T09:52:53Z'
+    projects:
+    - id: abcd1234cdef1234abcd1234cdef1234
+      name: My Android Project
+      main_format: xml
+      created_at: '2015-01-28T09:52:53Z'
+      updated_at: '2015-01-28T09:52:53Z'
--- a/doc/compiled.json
+++ b/doc/compiled.json
@@ -608,6 +608,72 @@
           "updated_at": "2015-01-28T09:52:53Z"
         }
       },
+      "custom_metadata_label": {
+        "type": "object",
+        "title": "custom_metadata_label",
+        "properties": {
+          "id": {
+            "type": "string"
+          },
+          "name": {
+            "type": "string"
+          },
+          "description": {
+            "type": "string"
+          },
+          "data_type": {
+            "type": "string",
+            "enum": [
+              "boolean",
+              "date",
+              "link",
+              "multi_select",
+              "number",
+              "single_select",
+              "text"
+            ]
+          },
+          "user": {
+            "$ref": "#/components/schemas/user_preview"
+          },
+          "projects": {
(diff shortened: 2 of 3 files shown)


jablan (changes requested):

paths/custom_metadata_labels/index.yaml
```diff
@@ -0,0 +1,53 @@
+---
+summary: List custom_metadata_labels
+description: List all custom_metadata_labels for an account.
+operationId: custom_metadata_labels/list
+tags:
+- CustomMetadataLabels
+parameters:
+- "$ref": "../../parameters.yaml#/X-PhraseApp-OTP"
+- "$ref": "../../parameters.yaml#/id"
```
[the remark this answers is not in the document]
jablan: I believe it should be `account_id`, probably similar to https://github.com/phrase/openapi/blob/b770c1bb352d2fc963954a27f20008f0016ec743/paths/glossaries/index.yaml#L9

paths/custom_metadata_labels/index.yaml
```diff
@@ -0,0 +1,82 @@
...
+  This endpoint is only available to accounts with advanced plans or above.
+operationId: custom_metadata_labels/list
+tags:
+- Custom Metadata
+parameters:
+- "$ref": "../../parameters.yaml#/X-PhraseApp-OTP"
+- "$ref": "../../parameters.yaml#/account_id"
+- "$ref": "../../parameters.yaml#/page"
+- "$ref": "../../parameters.yaml#/per_page"
+- description: specify the branch to use
```
jablan, line 15: how does the branch come into play here?

paths/custom_metadata_labels/index.yaml
```diff
@@ -0,0 +1,82 @@
...
+  example: my-feature-branch
+  name: branch
+  in: query
+  schema:
+    type: string
+- description: ids of projects that the labels belong to
```
jablan, line 21: so, if a project list is specified, that will constraint only to labels assigned to any? all? of these projects? and without it, it will show all the labels on an account?

paths/custom_metadata_labels/index.yaml
```diff
@@ -0,0 +1,82 @@
...
+  example:
+  - abcd1234cdef1234abcd1234cdef1234
+  name: project_ids
+  in: query
+  schema:
+    type: array
+    items:
+      type: string
+- description: 'Sort criteria. Can be one of: key_name, created_at, updated_at.'
```
jablan, line 30: `key_name` or `name`? should this also be an enum (as well as `order`)?

schemas/custom_metadata_label.yaml
```diff
@@ -0,0 +1,50 @@
...
+        - boolean
+        - date
+        - link
+        - multi_select
+        - number
+        - single_select
+        - text
+    user:
+      "$ref": "./user_preview.yaml#/user_preview"
+    account:
```
jablan, line 24: why an account per every label? shouldn't they all share the same account?

schemas/custom_metadata_label.yaml
```diff
@@ -0,0 +1,50 @@
...
+      type: string
+      enum:
+        - boolean
+        - date
+        - link
+        - multi_select
+        - number
+        - single_select
+        - text
+    user:
```
jablan, line 22: is this the user that created the particular label?

paths/custom_metadata_labels/index.yaml
jablan: is this filtering needed at this point? who requested for it?

PR #373, opened 2023-07-28
fix(schemas): Fix gitlab_sync type mismatch

### LINKED ISSUE #372: [BUG] phrase git_lab_sync history = ERROR: json: cannot unmarshal object into Go struct field GitlabSyncHistory.errors
Is exactly the same issue with #362, but with another field, it was not caught before, cause it was failing first at the status field

When executing the `phrase git_lab_sync history` in the latest version of the phrase-cli, it raises an error message:
```bash
ERROR: json: cannot unmarshal object into Go struct field GitlabSyncHistory.errors of type string
```
This seems to be related to a type mismatch between the model returned by the API and the model that the CLI is expecting 
[gitlab_sync_history.yaml#L10-L13](https://github.com/phrase/openapi/blob/master/schemas/gitlab_sync_history.yaml#L10-L13)
```yaml
errors:
   type: array
   items:
       type: string
```
but the response is in fact an array of object

```json
"errors":[
   {
      "error":"GitSync::Gitlab::Importer::UploadFailed",
      "message":"Unauthorized"
   }
],
```

I will once again create an MR to fix this

horstmannmat:
fixes #372 

Signed-off-by: Matheus Horstmann <>

### DIFF
--- a/doc/compiled.json
+++ b/doc/compiled.json
@@ -511,7 +511,15 @@
           "errors": {
             "type": "array",
             "items": {
-              "type": "string"
+              "type": "object",
+              "properties": {
+                "error": {
+                  "type": "string"
+                },
+                "message": {
+                  "type": "string"
+                }
+              }
             }
           },
           "date": {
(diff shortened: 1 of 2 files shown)


theSoenke:
Thanks a lot for addressing this too! I'll release a new version right away

theSoenke (approved):

PR #380, opened 2023-08-17
feat(TSE-950): Document comment_reactions endpoints

Varpusparvi:
https://github.com/phrase/phrase/pull/11577
https://memsourcegroup.atlassian.net/browse/TSE-950

### DIFF
--- a/main.yaml
+++ b/main.yaml
@@ -112,7 +112,7 @@ tags:
         </table>
       </div>
   - name: Comments
-  - name: Comment reactions
+  - name: Comment Reactions
   - name: Distributions
   - name: Documents
   - name: Formats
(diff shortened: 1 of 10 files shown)


jablan (changes requested):
I believe CLI version needs to be 2.9, as that's where this endpoint will be available.

main.yaml
```diff
@@ -275,6 +276,7 @@ x-tagGroups:
       - Organization Job Templates
       - Organization Job Template Locales
       - Comments
+      - Comment reactions
```
jablan, line 279: IIRC some generated classes are named after these tags. Not sure whether it is affected by the case of the tags, but looking that all other tags are capitalized, I suggest renaming this to `Comment Reactions`, with both words capitalized.

jablan (approved):

PR #383, opened 2023-08-23
feat(API): Introduce comment replies endpoints

Varpusparvi:
- Added comment replies endpoint docs
- Adjusted comment #create to include locale_ids
- Adjusted reaction #create to include example

https://github.com/phrase/phrase/pull/11643
https://memsourcegroup.atlassian.net/browse/TSE-869

### DIFF
--- a/paths/comment_reactions/create.yaml
+++ b/paths/comment_reactions/create.yaml
@@ -15,6 +15,12 @@ parameters:
   in: query
   schema:
     type: string
+- description: specify the emoji for the reaction
+  example: 👍
+  name: emoji
+  in: query
+  schema:
+    type: string
 responses:
   '201':
     description: Created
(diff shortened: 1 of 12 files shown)


paths/comment_reactions/create.yaml
```diff
@@ -43,14 +49,14 @@ x-code-samples:
...
+      -d '{"branch":"my-feature-branch", "emoji": "👍"}' \
       -H 'Content-Type: application/json'
 - lang: CLI v2
   source: |-
     phrase reactions create \
     --project_id <project_id> \
     --key_id <key_id> \
     --comment_id <comment_id> \
-    --branch my-feature-branch \
+    --data '{"branch":"my-feature-branch", "emoji": "👍"]}' \
```
jablan, line 60: interesting, does this mean that branch can't be specified with a dedicated switch?

jablan (approved):
Generally OK, just curious about the branch parameter

paths/comment_reactions/create.yaml
Varpusparvi: @jablan I'm exactly wondering about this too. In some places we have --branch param and in some just the --data with branch inside 🤔 What do you think? How should we do this so it will definitely work?

jablan: IMO it would be the best if CLI could have parameters specified through the dedicated switches, and translate that to JSON payload to the API, but not sure if that works and how.

PR #393, opened 2023-08-24
feat(API): Document new query parameters

Varpusparvi:
- Document new query parameters

https://github.com/phrase/phrase/pull/11643
https://memsourcegroup.atlassian.net/browse/TSE-869

### DIFF
--- a/paths/comment_replies/index.yaml
+++ b/paths/comment_replies/index.yaml
@@ -17,6 +17,20 @@ parameters:
   in: query
   schema:
     type: string
+- description: Search query for comment messages
+  example: Some comment content
+  name: query
+  in: query
+  schema:
+    type: string
+- description: Specify the filter for the comments
+  name: filters
+  in: query
+  schema:
+    type: array
+    example: [read, unread]
+    items:
+      type: string
 responses:
   '200':
     description: OK
@@ -49,7 +63,7 @@ x-code-samples:
     curl "https://api.phrase.com/v2/projects/:project_id/keys/:key_id/comments/:comment_id/replies" \
       -u USERNAME_OR_ACCESS_TOKEN \
       -X GET \
-      -d '{"branch":"my-feature-branch"}' \
+      -d '{"branch":"my-feature-branch", "filters": ["unread", "read"], "query": "Some comment content"}' \
       -H 'Content-Type: application/json'
 - lang: CLI v2
   source: |-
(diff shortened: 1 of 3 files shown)


Varpusparvi:
I will try to bump the GO version on the generated CLI PR @jablan

paths/comments/index.yaml
```diff
@@ -44,12 +66,43 @@ x-code-samples:
...
+      -X GET \
+      -d '{"branch":"my-feature-branch", "filters": ["unread", "read"], "query": "Some comment content", "locale_ids": ["someId"] }' \
+      -H 'Content-Type: application/json'
 - lang: CLI v2
   source: |-
     phrase comments list \
     --project_id <project_id> \
     --key_id <key_id> \
-    --branch my-feature-branch \
+    --data '{"branch":"my-feature-branch", "filters": ["unread", "read"], "query": "Some comment content", "locale_ids": ["someId"] }' \
```
jablan, line 78: I believe this is a GET request, so the parameters should be passed as query params, not as JSON?

Varpusparvi: Adjusted. This is probably something we have to check for every endpoint also, seems like it's pretty mixed atm 🤔

paths/comment_replies/index.yaml
```diff
@@ -49,14 +63,38 @@ x-code-samples:
     curl "https://api.phrase.com/v2/projects/:project_id/keys/:key_id/comments/:comment_id/replies" \
       -u USERNAME_OR_ACCESS_TOKEN \
       -X GET \
-      -d '{"branch":"my-feature-branch"}' \
+      -d '{"branch":"my-feature-branch", "filters": ["unread", "read"], "query": "Some comment content"}' \
```
jablan, line 66: this should go into query I believe

paths/comments/index.yaml
```diff
@@ -44,12 +66,46 @@ x-code-samples:
 - lang: Curl
   source: |-
     curl "https://api.phrase.com/v2/projects/:project_id/keys/:key_id/comments?branch=my-feature-branch" \
-      -u USERNAME_OR_ACCESS_TOKEN
+      -u USERNAME_OR_ACCESS_TOKEN \
+      -X GET \
+      -d '{"branch":"my-feature-branch", "filters": ["unread", "read"], "query": "Some comment content", "locale_ids": ["someId"] }' \
```
jablan, line 71: same here

jablan (changes requested):
the curl examples seem wrong, with parameters in a `data` payload rather than in the query. have you tried these examples against the app, I'm curious whether both ways of passing them work?

jablan (approved):

PR #404, opened 2023-08-28
docs: Mention passing name instead of locale ID

jablan:


theSoenke (commented):
Nice, to document this. Just `locale_id_as_id` is not entirely clear to me. Should it maybe be something like `locale_id_or_name` or `locale_id_name`?

jablan:
@theSoenke `_as_id` suggests that the identifier is passed as `id` (and not as `locale_id`). There are other endpoints that would benefit from a similar clarification, and they would have `project_id_as_id` or `account_id_as_id` for example.

theSoenke (approved):
Nice, improvement. Ideally we would use `locale_id` as part of the path I think. Then we could just use the existing `locale_id` param and the same for all other places where we use `{id}` in the path. But as we discussed it's not clear yet what side effects this would have

PR #412, opened 2023-09-12
feat: Optionally tag only affected keys on upload [TSI-2041]

jablan:


paths/uploads/create.yaml
```diff
@@ -124,4 +124,9 @@ requestBody:
             description: Indicated whether the imported translations should be marked as reviewed. This setting is available if the review workflow is enabled for the project.
             type: boolean
             example:
+          tag_only_affected_keys:
+            description: Indicates whether only keys affected (created or updated) by the upload should be tagged. The default is `false`
+            type: boolean
+            default: false
+            example:
 x-cli-version: '2.5'
```
theSoenke, line 132: I think we needed to bump this CLI version also for new params, right?

theSoenke (approved):

PR #415, opened 2023-09-27
feat(API): Implement figma attachments endpoints

Varpusparvi:
Screenshot 2023-09-27 at 14 29 21

- https://phrase.atlassian.net/browse/TSE-613
- https://github.com/phrase/phrase/pull/11856

### DIFF
--- a/paths/figma_attachment_keys/create.yaml
+++ b/paths/figma_attachment_keys/create.yaml
@@ -1,0 +1,44 @@
+---
+summary: Attach the Figma attachment to a key
+description: Attach the Figma attachment to a key
+operationId: figma_attachment_key/create
+tags:
+  - Key's Figma attachments
+parameters:
+  - "$ref": "../../parameters.yaml#/X-PhraseApp-OTP"
+  - "$ref": "../../parameters.yaml#/project_id"
+  - "$ref": "../../parameters.yaml#/figma_attachment_id"
+  - "$ref": "../../parameters.yaml#/id"
+  - description: specify the branch to use
+    example: my-feature-branch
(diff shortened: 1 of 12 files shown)


Varpusparvi:
Reminder to self: I will bump GO before Bumping the CLI

paths/figma_attachment_keys/create.yaml
```diff
@@ -0,0 +1,44 @@
+---
+summary: Attach the Figma attachment to a key
+description: Attach the Figma attachment to a key
+operationId: figma_attachment_key/create
```
jablan, line 4: wouldn't it be more logical to add these operations to `figma_attachments` (or perhaps `keys`) instead?

Varpusparvi: Adjusted, not exactly sure if this is what you meant 
[CR](https://github.com/phrase/openapi/pull/415/commits/f4872f5e44bec82004dfae7f99b2d3260737a6fb)

jablan: yeah, but they'd also have to live on `paths/figma_attachments` IMO

jablan: but I guess the API PR has already merged in phrase?

Varpusparvi: Yes the API PR is merged. Can we have two `create.yaml`/`destroy.yaml` in the same folder? I could adjust this. On the API PR we agreed to not use custom named methods and instead went with second controller to have option for second create/destroy paths

jablan (approved):

PR #417, opened 2023-10-02
docs: Expand multiline descriptions by default [TSI-2024]

jablan:
Avoid confusing users with descriptions collapsed by default and not being obvious that they can be expanded

https://phrase.atlassian.net/browse/TSI-2024


theSoenke (approved):
wow, also totally forgot there was this expand action. way better to have it extended by default

PR #418, opened 2023-10-05
feat: Update openapi-generator to v7

jablan:
Big bump for openapi-generator from v4.3 to v7. Tried to keep the generated code as close as possible, by modifying the templates which have previously been dumped to `openapi-generator/templates` folder.

For comparing results, I was checking out another copy of this project, generating the libraries, and then running a recursive folder comparison tool.


theSoenke:
@jablan great we could keep the existing interfaces with the latest generator 🔥 do you think it would make sense to add those tests first before we merge this to check that they pass before and after?

jablan:
@theSoenke that'd be great, if can be done in this sprint.

hahmed-dev (approved):
Looks good for the most part! Great Job!!

theSoenke (approved):
🚀 also want to bump the library versions in this PR already?

PR #420, opened 2023-10-10
test: Add Java test [TSI-2113]

jablan:
Adds basic test for Java library
https://phrase.atlassian.net/browse/TSI-2113


jablan:
@theSoenke  I was looking into moving `.gitignore` chunks into appropriate subdirectories, but they already contain generated `.gitignore` files, so we'd have to customize corresponding mustache templates, which felt a bit wrong. also, having all these in a single place makes more sense IMO.

theSoenke (approved):
cool, another one done

PR #426, opened 2023-10-12
feat: [TSI-2083] enable format_options argument for java-client  

hahmed-dev:
## Assumptions
Some assumptions made while working on the PR:

- `format_options` argument will be key-value based map with values mostly being of primitive types
- Initially explored the possibility of using params with `style` `deepObject` and `explode: true` option as suggested [here](https://stackoverflow.com/questions/48491688/how-to-define-parameters-with-square-brackets-in-openapi-swagger/48498055#48498055) as a fix but subsquently wasn't able to find appropriate mustache directive to determine if the param has any of these attributes (deepObject, explode: true) which led to adding a fix by checking if input is of the type Map and if yes, manually convert it to key value based `Pair` objects.

## Problem Statement
Phrase API accepts `format_options` argument on [locales download](https://developers.phrase.com/api/#get-/projects/-project_id-/locales/-id-/download) endpoint. 

For java-client the format_option argument was not being properly parsed and embedded in the query params resulting in a request url like below

```
https://api.phrase.com/v2/projects/e4c0450ccc8dcc087fac8351ecc89eb6/locales/01877decc84124a75fdd0a4fc147d8de/download?file_format=properties&tags=menu&format_options=%7B%22omit_separator_space%22%3Atrue%7D&encoding=UTF-8
```

## Solution
This pull requests adds the ability to parse any query params of type map and embed them to the URL's query string as proper key value pairs resulting in request URLs like:
```
localhost:3000/api/v2/projects/:project_id/locales/:id/download?file_format=i18next_4&format_options%5Bnesting%5D=false
```

### DIFF
--- a/clients/java/src/test/java/com/phrase/client/api/LocalesApiTest.java
+++ b/clients/java/src/test/java/com/phrase/client/api/LocalesApiTest.java
@@ -137,9 +137,14 @@ public class LocalesApiTest {
 
         mockBackend.enqueue(mockResponse);
 
-        Map<String, String> formatOptionsMap = new HashMap<>();
+        
+        Map<String, String> nestedFormatOptionsMap = new HashMap<>();
+        nestedFormatOptionsMap.put("nested_option", "sub_option");
+        
+        Map<String, Object> formatOptionsMap = new HashMap<>();
         formatOptionsMap.put("omit_separator_space", "true");
         formatOptionsMap.put("fallback_language", "en");
+        formatOptionsMap.put("more_options", nestedFormatOptionsMap);
 
         String projectId = "MY_PROJECT_ID";
         String id = "MY_ID";
@@ -168,7 +173,8 @@ public class LocalesApiTest {
         Assert.assertEquals("Correct file contents", fileContents, body);
 
         RecordedRequest recordedRequest = mockBackend.takeRequest();
-        Assert.assertEquals("Request path", "//projects/MY_PROJECT_ID/locales/MY_ID/download?format_options%5Bomit_separator_space%5D=true&format_options%5Bfallback_language%5D=en", recordedRequest.getPath());
+        // for some reason with deep nested query params, ordering of query params change
+        Assert.assertEquals("Request path", "//projects/MY_PROJECT_ID/locales/MY_ID/download?format_options%5Bomit_separator_space%5D=true&format_options%5Bmore_options%5D%5Bnested_option%5D=sub_option&format_options%5Bfallback_language%5D=en", recordedRequest.getPath());
     }
 
     /**
--- a/clients/java/src/test/java/com/phrase/client/api/LocalesApiTest.java
+++ b/clients/java/src/test/java/com/phrase/client/api/LocalesApiTest.java
@@ -137,6 +137,10 @@ public class LocalesApiTest {
 
         mockBackend.enqueue(mockResponse);
 
+        Map<String, String> formatOptionsMap = new HashMap<>();
+        formatOptionsMap.put("omit_separator_space", "true");
+        formatOptionsMap.put("fallback_language", "en");
+
         String projectId = "MY_PROJECT_ID";
         String id = "MY_ID";
         String xPhraseAppOTP = null;
@@ -151,7 +155,7 @@ public class LocalesApiTest {
         Boolean includeTranslatedKeys = null;
         Boolean keepNotranslateTags = null;
         Boolean convertEmoji = null;
-        Object formatOptions = null;
+        Object formatOptions = formatOptionsMap;
         String encoding = null;
         Boolean skipUnverifiedTranslations = null;
         Boolean includeUnverifiedTranslations = null;
@@ -164,7 +168,7 @@ public class LocalesApiTest {
         Assert.assertEquals("Correct file contents", fileContents, body);
 
         RecordedRequest recordedRequest = mockBackend.takeRequest();
-        Assert.assertEquals("Request path", "//projects/MY_PROJECT_ID/locales/MY_ID/download", recordedRequest.getPath());
+        Assert.assertEquals("Request path", "//projects/MY_PROJECT_ID/locales/MY_ID/download?format_options[omit_separator_space]=true&format_options[fallback_language]=en", recordedRequest.getPath());
     }
 
     /**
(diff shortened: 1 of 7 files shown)


doc/compiled.json
```diff
@@ -6908,7 +6908,9 @@
             "schema": {
               "type": "object",
               "properties": {}
-            }
+            },
+            "style": "deepObject",
+            "explode": true
```
hahmed-dev, line 6913: I checked locally and without these options as well, the test cases pass. Can be removed unless we decide to make any alterations or changes to the currently proposed solution.

paths/locales/download.yaml
```diff
@@ -73,6 +73,8 @@ parameters:
     schema:
       type: object
       properties: {}
+    style: deepObject
+    explode: true
```
hahmed-dev, line 77: This can be removed? The current solution does not rely on these properties, tested locally.

openapi-generator/templates/java/libraries/okhttp-gson/api.mustache
```diff
@@ -100,7 +100,12 @@ public class {{classname}} {
         {{javaUtilPrefix}}List<Pair> localVarCollectionQueryParams = new {{javaUtilPrefix}}ArrayList<Pair>();
         {{#queryParams}}
         if ({{paramName}} != null) {
+            {{#isMap}}
```
jablan, line 103: I believe you should use `#isDeepObject` instead. You can see it's available to the API template if you run generate with `--global-property debugOperations=true` (also `isExplode` is available)

openapi-generator/templates/java/libraries/okhttp-gson/ApiClient.mustache
```diff
@@ -654,6 +654,26 @@ public class ApiClient {
         }
     }
 
+    public List<Pair> mappedParameterToPairs(String name, Object parameter){
+        List<Pair> params = new ArrayList<Pair>();
+
+        if(parameter instanceof Map){
```
jablan, line 660: there's additional problem that we'll have to address at some point: these objects can be nested. so, it should be possible to generate a parameter such as `formatOptions[customMetadataColumns][order]=B` and so on :smile:  Perhaps we could address that in a separate ticket

openapi-generator/templates/java/libraries/okhttp-gson/api.mustache
hahmed-dev: Thanks, will take a look, the document it printed out for me upon running with this argument,

- niether included any documentations for get end points
- nor `isDeepObject` or `isExplode` properties

But let me check again, would be more specific

jablan: or, you can pass `{{style}}` as a parameter to `mappedParameterToPairs`, and then handle it appropriately (leaving other styles not covered until we need them).

jablan: note that there are different `debug...` flags possible to pass as `global-property`. https://openapi-generator.tech/docs/debugging/#templates

openapi-generator/templates/java/libraries/okhttp-gson/ApiClient.mustache
hahmed-dev: I guess in case if a value against format_option is an array, it should already be [handled](https://github.com/phrase/openapi/blob/master/openapi-generator/templates/java/libraries/okhttp-gson/ApiClient.mustache#L643)? If not I will check and see if that can be addressed as well.

openapi-generator/templates/java/libraries/okhttp-gson/api.mustache
hahmed-dev: Will check this. Thanks for bringing it up.

openapi-generator/templates/java/libraries/okhttp-gson/ApiClient.mustache
jablan: seems it covers the basic case, but I guess arrays can also be nested :grimacing: never mind, we can cross that bridge when we get there

hahmed-dev: 😱

jablan: Here's something that we can use as a reference implementation: https://github.com/rails/rails/blob/main/activesupport/lib/active_support/core_ext/object/to_query.rb but I suggest leaving that for another ticket.

hahmed-dev: Was already looking into nested arguments, that part is addressed with the latest commit. Regarding values being arrays/nested array, can see if it can be handled separately.

hahmed-dev: @theSoenke by same issue do you mean, it will require same patch like this?? For map based query params?

clients/java/src/test/java/com/phrase/client/api/LocalesApiTest.java
```diff
@@ -168,7 +173,8 @@ public void localeDownloadTest() throws ApiException, IOException, InterruptedEx
         Assert.assertEquals("Correct file contents", fileContents, body);
 
         RecordedRequest recordedRequest = mockBackend.takeRequest();
-        Assert.assertEquals("Request path", "//projects/MY_PROJECT_ID/locales/MY_ID/download?format_options%5Bomit_separator_space%5D=true&format_options%5Bfallback_language%5D=en", recordedRequest.getPath());
+        // for some reason with deep nested query params, ordering of query params change
```
jablan, line 176: rails implementation simply sorts the params by name at end. :)

jablan (approved):
nice!

openapi-generator/templates/java/libraries/okhttp-gson/ApiClient.mustache
theSoenke: @hahmed-dev `format_options` should already work. But for more deeply nested params it will need a similar fix I think

openapi-generator/templates/java/libraries/okhttp-gson/api.mustache
hahmed-dev: Thanks Mladen, using `debugOperations` was able to find the name of the directive.

docstun:
@hahmed-dev is this still valid?

hahmed-dev:
@docstun yes, we initially were waiting to merge the open api version bump PR since its based out of that, but we discussed last week, there is no dependency and we can merge this independently as well. I will try and shortly open a new PR and if that gets merged close this one.

jablan (approved):

PR #427, opened 2023-10-12
test: Add Python tests

theSoenke:
Add basic tests for Python


.gitignore
```diff
@@ -1,5 +1,8 @@
 clients/go/
-clients/python/
+
+clients/python/*
+!clients/python/test/test_locales_api.py
```
jablan, line 4: are you sure that this works, as one can't exclude a single file when the whole directory has been ignored.

>  It is not possible to re-include a file if a parent directory of that file is excluded.

https://git-scm.com/docs/gitignore

jablan (approved):

.gitignore
theSoenke: You're right. Doesn't make a difference at all. Then the question is whether we should explicitly exclude all other files or we perform `git add -f` for files we want to track in the future. In both cases we manually we to add the .gitignore entry or run the command

jablan: hm somehow I thought that git won't recognize changes in a file even if it's force-added, but it doesn't seem to be the case. in that case, I think it's easier to do `git add -f` whenever we add a new file that needs to be preserved, rather than maintaining this weird structure in `.gitignore`.

PR #436, opened 2023-10-13
fix(python): pass format_options correctly

theSoenke:
Fix passing nested params in GET requests


clients/python/test/test_locales_api.py
```diff
@@ -54,12 +54,34 @@ def test_locale_delete(self):
...
+        mock_get.return_value.getheader.side_effect = { 'Content-Disposition': None }.get
+
+        with phrase_api.ApiClient(self.configuration) as api_client:
+            api_instance = phrase_api.api.locales_api.LocalesApi(api_client)
+            api_response = api_instance.locale_download("project_id_example", "en", file_format="simple_json", format_options={"enable_pluralization": True, "custom_metadata_columns": { "E": "text" }})
+
+
+            self.assertEqual("https://api.phrase.com/v2/projects/project_id_example/locales/en/download", mock_get.call_args_list[0].args[1])
+
+            self.assertEqual([('file_format', 'simple_json'), ('format_options[enable_pluralization]', True), ('format_options[custom_metadata_columns][E]', 'text')], mock_get.call_args_list[0][1]['fields'])
```
jablan, line 78: is there a way to assert the final URL?

openapi-generator/templates/python/rest.mustache
```diff
@@ -152,6 +152,9 @@ class RESTClientObject(object):
         if 'Content-Type' not in headers:
             headers['Content-Type'] = 'application/json'
 
+        if query_params:
+            query_params = self.flatten_params(query_params)
```
jablan, line 156: I guess this works too, but IMO the proper way would be to do it similar to `parameters_to_tuples` in https://github.com/phrase/openapi/blob/tsi-1763-python-format-options/openapi-generator/templates/python/api_client.mustache#L164 and respecting given style (`deepObject` in our case)

clients/python/test/test_locales_api.py
theSoenke: Found a solution by mocking `urlopen`

openapi-generator/templates/python/rest.mustache
theSoenke: That seems to be the better place. But regarding `deepObject` what would you expect there? I don't have this parameter information there anymore. Shouldn't it be enough here to flatten all dicts?

theSoenke: Of course. Really thought I've removed these lines before 👀

jablan (approved):

PR #440, opened 2023-10-20
test: Add Go tests

hahmed-dev:
Add basic tests for Go Lang.

Related Issue: https://phrase.atlassian.net/browse/TSI-2123


Makefile
```diff
@@ -33,8 +33,6 @@ ruby:
 	openapi-generator-cli generate -i tmp/compiled.yaml -g ruby -o clients/ruby -c ./openapi-generator/ruby_lang.yaml
 go:
 	openapi-generator-cli generate -i tmp/compiled.yaml -g go -o clients/go -c ./openapi-generator/go_lang.yaml
-	go install golang.org/x/tools/cmd/goimports@latest
-	goimports -w clients/go
```
hahmed-dev, line 37: Was getting the [following](https://github.com/phrase/openapi/actions/runs/6586658228/job/17895395665#step:3:578) issue with these two lines, looked for a fix which suggested updating system's bashprofile. Tested with remove them, apparently it works, looked up briefly what [goimport](https://pkg.go.dev/golang.org/x/tools/cmd/goimports) does, apparently it is some tidying up around imports.

hahmed-dev: Might be needed after all, fixing it.

.gitignore
```diff
@@ -56,6 +54,19 @@ clients/php/test/Api/*
 !clients/php/test/Api/LocalesApiTest.php
 !clients/php/test/Api/UploadsApiTest.php
 
+clients/go/*.go
```
jablan, line 57: We've discussed this here https://github.com/phrase/openapi/pull/427#discussion_r1356900271 and decided not to go this individual exclusion route. so, feel free to leave this file as is, and only `git add -f` files that need to be added.

clients/go/test/api_locales_test.go
```diff
@@ -0,0 +1,157 @@
...
+	}))
+
+	defer server.Close()
+
+	configuration := phrase.NewConfiguration()
+	configuration.BasePath = server.URL
+	apiClient := phrase.NewAPIClient(configuration)
+	
+
+	t.Run("Test LocalesApiService AccountLocales", func(t *testing.T) {
```
jablan, line 41: I'd delete these unused tests.

clients/go/test/api_locales_test.go
```diff
@@ -0,0 +1,157 @@
...
+		require.NotNil(t, resp)
+		assert.Equal(t, 200, httpRes.StatusCode)
+		assert.Equal(t, "/projects/project_id_example/locales/locale_id", requestUrl.Path)
+		assert.Equal(t, "PATCH", httpRes.Request.Method)
+	})
+
+	t.Run("Test LocalesApiService LocalesList", func(t *testing.T) {
+		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// Send the mock response
+			response := `[{"foo": "bar"}]`
```
jablan, line 132: can we also return a proper format and actually assert that returned locales are deserialized properly?

clients/go/test/api_uploads_test.go
```diff
@@ -0,0 +1,77 @@
...
+	}))
+
+	defer server.Close()
+
+	configuration := phrase.NewConfiguration()
+	configuration.BasePath = server.URL
+	apiClient := phrase.NewAPIClient(configuration)
+
+	t.Run("Test UploadsApiService UploadCreate", func(t *testing.T) {
+		localVarOptionals := phrase.UploadCreateOpts{}
```
jablan, line 40: not sure how easy would it be to pass an actual file to this call and then assert that the body contains it?

jablan (dismissed):
looks good, some questions added, but nothing super important.

clients/go/test/api_uploads_test.go
hahmed-dev: Looked into it, tried to access request via 

```
httpRes.Request.Body which but it returns {}
```

Also tried to explore possibility to read the request Body inside the [mock server](https://github.com/phrase/openapi/pull/440/files#diff-8208c8a7de5d4c0b513c42d411c1f576a37ec316caa3d0dc4dccbed902cc150dR27) but I will have to look further how to handle the returned Body object of type `io.ReadCloser` which by default is unreadable. If its important, I can further investigate if its possible.

theSoenke (approved):
LGTM ✅

PR #463, opened 2023-11-03
feat(CLI): On pull error, print the API response [TSI-1750]

jablan:
Now when there's an error on CLI pull, the users are left in dark as the server response (which usually explains the reason for the error) is not printed out. This PR addresses that.

https://phrase.atlassian.net/browse/TSI-1750

Related: https://github.com/phrase/phrase-cli/issues/126


theSoenke (approved):
Really helpful

PR #474, opened 2023-11-30
feat(API): add Custom Metadata endpoints

anthonyshchang:
https://github.com/phrase/openapi/assets/108525179/2a4151e2-f39c-49de-bf1d-b408d19ed7f6

### DONT MERGE
DONT MERGE UNTIL FEATURE TOGGLE IS REMOVED!

### Ticket
https://phrase.atlassian.net/browse/TSS-2511


forelabs (approved):

parameters.yaml
```diff
@@ -31,6 +31,14 @@ account_id:
   required: true
   schema:
     type: string
+data_type:
```
jablan, line 34: I suggest renaming this to `custom_metadata_data_type` or `custom_metadata_type` to avoid confusion

paths.yaml
```diff
@@ -47,6 +47,18 @@
     "$ref": "./paths/styleguides/update.yaml"
   delete:
     "$ref": "./paths/styleguides/destroy.yaml"
+"/accounts/{account_id}/custom_metadata/labels":
+  get:
+    "$ref": "./paths/custom_metadata_properties/index.yaml"
+"/accounts/{account_id}/custom_metadata/labels/{id}":
+  get:
+    "$ref": "./paths/custom_metadata_properties/show.yaml"
+  post:
```
jablan, line 56: shouldn't POST be on a collection rather than on an item

paths/custom_metadata_properties/create.yaml
```diff
@@ -0,0 +1,83 @@
+---
+summary: Create a property
+description: Create a new custom metadata property.
+operationId: custom_metadata_property/create
+tags:
+- Custom Metadata
+parameters:
+- "$ref": "../../parameters.yaml#/X-PhraseApp-OTP"
+- "$ref": "../../parameters.yaml#/account_id"
+- "$ref": "../../parameters.yaml#/id"
```
jablan, line 10: see above comment

paths.yaml
```diff
@@ -47,6 +47,18 @@
     "$ref": "./paths/styleguides/update.yaml"
   delete:
     "$ref": "./paths/styleguides/destroy.yaml"
+"/accounts/{account_id}/custom_metadata/labels":
```
jablan, line 50: so we decided to leave `labels` in the API rather than renaming to `properties`?

jablan (commented):
I had some questions

paths/custom_metadata_properties/index.yaml
```diff
@@ -10,7 +10,7 @@ tags:
 parameters:
 - "$ref": "../../parameters.yaml#/X-PhraseApp-OTP"
 - "$ref": "../../parameters.yaml#/account_id"
-- "$ref": "../../parameters.yaml#/data_type"
+- "$ref": "../../parameters.yaml#/custom_metadata_data_type"
```
jablan, line 13: I'm curious, why is there a data_type parameter? is it for filtering properties only for the given type? what was the use case for adding this?

paths/custom_metadata_properties/index.yaml
```diff
@@ -10,7 +10,7 @@ tags:
...
 - description: id of project that the properties belong to
   example:
   - abcd1234cdef1234abcd1234cdef1234
```
jablan, line 16: why an array here?

jablan: yeah, scalar would just be

```yaml
  example: abcd1234cdef1234abcd1234cdef1234
```

paths.yaml
jablan: so now needs to be updated here too, right?

malvarezphrase (approved):

ildarsafin:
Tested the creation of key and response body: We are getting CM inside:

Screenshot 2023-12-13 at 10 25 18

parameters.yaml
```diff
@@ -35,9 +35,18 @@ custom_metadata_data_type:
   in: query
   name: data_type
   description: Data Type of Custom Metadata Property
-  required: false
+  required: true
```
jablan, line 38: by doing this, it is also mandatory in the `index` call, right? I'm pretty sure we don't want that.

ildarsafin: this is actually confusing since it's rendered also in create/update CM pages and it's shown as not required which is not true

jablan: yeah, I'm afraid it can't be reused as a single reference then.

jablan: wait, perhaps it can be specified using `required` array

ildarsafin: which one?

jablan: nah, forget it. you can't reuse the same reference.

PR #491, opened 2023-12-14
feat(CLI): Improve push error display [TSI-1736]

jablan:
Several smaller improvements on push errors:
* Print error response from server on upload create
* Print link to the upload upon upload processing error (the response does not contain more info)
* Return non-zero exit code when any of the uploads were not processed successfully (https://github.com/phrase/phrase-cli/issues/135)

### DIFF
--- a/clients/cli/cmd/internal/push.go
+++ b/clients/cli/cmd/internal/push.go
@@ -3,6 +3,7 @@ package internal
 import (
 	"bufio"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


clients/cli/cmd/internal/push.go
```diff
@@ -202,6 +209,9 @@ func (source *Source) Push(client *phrase.APIClient, waitForResults bool, branch
 			fmt.Fprintln(os.Stderr, strings.Repeat("-", 10))
 		}
 	}
+	if !noErrors {
+		return errors.New("not_all_files_processed")
```
theSoenke, line 213: Maybe instead of an error code like looking message we could write a short sentence what happened?

jablan: :heavy_check_mark: following other cases in the codebase, I rewrote this as a (semi)sentence

theSoenke (approved):

PR #498, opened 2023-12-18
fix(GO): Fix format_option serialization when creating upload

hahmed-dev:
Format Options with nested properties/map was not being properly serialized in the upload creation request.
This PR adds a custom method to serialize the nested format_options appropriately as a part of the request.

Related Issue: https://phrase.atlassian.net/browse/TSI-2192

### DIFF
--- a/clients/go/test/api_uploads_test.go
+++ b/clients/go/test/api_uploads_test.go
@@ -17,7 +17,7 @@ import (
 	"testing"
 
 	"github.com/antihax/optional"
-	"github.com/phrase/phrase-go"
+	"github.com/phrase/phrase-go/v2"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
--- a/clients/go/test/api_uploads_test.go
+++ b/clients/go/test/api_uploads_test.go
@@ -23,28 +23,46 @@ import (
 )
 
 func Test_phrase_UploadsApiService(t *testing.T) {
-	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		// Send the mock response
-		response := `{"id": "1", "filename": "test.json", "format": "json", "state": "valid" }`
-		w.Header().Set("Content-Type", "application/json")
-		w.WriteHeader(http.StatusOK)
-
-		w.Write([]byte(response))
-	}))
-
-	defer server.Close()
-
-	configuration := phrase.NewConfiguration()
-	configuration.BasePath = server.URL
-	apiClient := phrase.NewAPIClient(configuration)
-
 	t.Run("Test UploadsApiService UploadCreate", func(t *testing.T) {
+		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// assert request body for proper serialization of tested format_options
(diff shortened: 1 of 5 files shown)


clients/go/test/api_uploads_test.go
```diff
@@ -17,34 +17,52 @@ import (
...
+
 		file, _ := os.Create("testfile.json")
 		fileFormat := optional.NewString("json")
 		fileObject := optional.NewInterface(file)
 		localeId := optional.NewString("99")
 
-		localVarOptionals := phrase.UploadCreateOpts{FileFormat: fileFormat, File: fileObject, LocaleId: localeId }
+		// setting format_options
+		formatOptions := make(map[string]interface{})
+		nestedOptions := make(map[string]interface{})
```
hahmed-dev, line 55: I have assumed, the inner map, if provided will be also of type `map[string]interface{}`. Allowing for values of different types.

openapi-generator/templates/go/client.mustache
```diff
@@ -163,6 +163,19 @@ func parameterToJson(obj interface{}) (string, error) {
 	return string(jsonBuf), err
 }
 
+// helper for serializing and setting mappeed parameters to query string
```
jablan, line 166: small typo in the comment

clients/go/test/api_uploads_test.go
```diff
@@ -17,34 +17,52 @@ import (
...
+		nestedOptions["de"] = "e"
+
+		formatOptions["top_level_key"] = 100
+		formatOptions["translation_columns"] = nestedOptions
+
+		formatOptionsMap := optional.NewInterface(formatOptions)
+
+		localVarOptionals := phrase.UploadCreateOpts{FileFormat: fileFormat, File: fileObject, LocaleId: localeId, FormatOptions: formatOptionsMap }
 		resp, httpRes, err := apiClient.UploadsApi.UploadCreate(context.Background(), "project_id", &localVarOptionals)
 		requestUrl := httpRes.Request.URL
```
jablan, line 67: I think the request content can be also asserted here instead of in the mock server handler. could be cleaner that way, rather having assertions all over the place.

jablan (approved):
couple of small remarks

clients/go/test/api_uploads_test.go
hahmed-dev: Very valid point, actually by the point request is sent, the request's body is no longer available, quoting go's doc

> // Request is the request that was sent to obtain this Response.
	// Request's Body is nil (having already been consumed).
	// This is only populated for Client requests.
	Request *Request

hahmed-dev: Let me still check if there is a better way to achieve this.

hahmed-dev: I tried to look into a more readable alternative. There is another option to use [responseRecorder](https://pkg.go.dev/net/http/httptest#ResponseRecorder) but looking at some examples I think its a similar approach where assertions will be made inside the handler function that responsible to stub the request.

jablan (approved):

clients/go/test/api_uploads_test.go
jablan: it's fine as is, no worries.

PR #499, opened 2023-12-19
fix(API): Create Custom Metadata endpoint fix [TSI-2222]

jablan:
Update Custom Metadata non-GET endpoints so that they comply with our standard way of passing parameters.

### DIFF
--- a/clients/go/test/api_custom_metadata_test.go
+++ b/clients/go/test/api_custom_metadata_test.go
@@ -2,11 +2,11 @@ package phrase
 
 import (
 	"context"
+	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"testing"
 
-	"github.com/antihax/optional"
 	"github.com/phrase/phrase-go/v2"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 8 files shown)


clients/go/test/api_custom_metadata_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+	t.Run("Test CustomMetadataService CustomMetadataCreate", func(t *testing.T) {
+		customMetadataPropertiesCreateParameters := phrase.CustomMetadataPropertiesCreateParameters{
+			Name:        "my_property",
+			DataType:    phrase.STRING,
+			Description: "my description",
+			ProjectIds:  []string{"project_id", "project_id2"},
+		}
+		localVarOptionals := phrase.CustomMetadataPropertyCreateOpts{}
+		resp, httpRes, err := apiClient.CustomMetadataApi.CustomMetadataPropertyCreate(context.Background(), "account_id", customMetadataPropertiesCreateParameters, &localVarOptionals)
```
hahmed-dev, line 46: Just one question passing `customMetadataPropertiesCreateParameters` here is equivalent to passing these arguments individually 

```
(ctx _context.Context, accountId string, name string, dataType CustomMetadataDataType
```

in the actual api `/clients/go/api_custom_metadata.go`?

jablan: no, these individual arguments are replaced by this PR with `CustomMetadataPropertiesCreateParameters` struct. technically, it's a breaking change, but I don't think anyone is using this yet.

paths/custom_metadata_properties/update.yaml
```diff
@@ -8,40 +8,40 @@ parameters:
...
-  in: query
-  schema:
-    type: array
-    items:
-      type: string
+requestBody:
+  required: true
+  content:
+    application/json:
+      schema:
```
hahmed-dev, line 15: Should we also add the required `name` and `datatype` section?
```suggestion
      schema:
        required:
        - name
        - data_type
```

jablan: yeah, I haven't checked the backend code, but I would guess that the data type is not changeable once the property is created, and the name is certainly not required, if one doesn't want to change it.

hahmed-dev (approved):

PR #518, opened 2024-01-05
feat(CLI): Optional cleanup after push [TSI-2234]

jablan:
Addresses https://github.com/phrase/phrase-cli/issues/22

We could reuse existing `UploadCleanup` command. The tricky part is that a single `push` consists of many uploads, each might contain a different set of keys. So, in order to properly cleanup unmentioned keys, we need to find an intersection of unmentioned keys for all uploads, and clean up only these. So `UploadCleanup` is changed to accept an array of upload IDs, and getting the list of unmentioned keys (in batches) is now decoupled by deleting unmentioned keys (also in batches), with calculating intersection in between.


clients/cli/cmd/internal/push.go
```diff
@@ -209,13 +212,29 @@ func (source *Source) Push(client *phrase.APIClient, waitForResults bool, branch
 			fmt.Fprintln(os.Stderr, strings.Repeat("-", 10))
 		}
 	}
-	if !noErrors {
+	if noErrors {
+		if waitForResults && cleanup {
```
theSoenke, line 216: Maybe you can show an error when the `cleanup` is passed but `wait` is missing? Otherwise it might cause some confusion why the cleanup didn't happen

clients/cli/cmd/internal/push.go
```diff
@@ -209,13 +212,29 @@ func (source *Source) Push(client *phrase.APIClient, waitForResults bool, branch
...
+		}
+	} else {
 		return errors.New("not all files were uploaded successfully")
 	}
 
 	return nil
 }
 
+func cleanupAfterUpload(client *phrase.APIClient, projectId string, uploadIds []string) error {
+	cleanupCommand := &UploadCleanupCommand{
```
theSoenke, line 227: How about putting the logic for deleting the keys into this method and then call it from the clean up command? Simulating a command to cleanup the keys feels a bit like the wrong way around

clients/cli/cmd/internal/upload_cleanup.go
```diff
@@ -30,52 +31,96 @@ func (cmd *UploadCleanupCommand) Run() error {
...
-		Q:       optional.NewString(q),
-		Branch:  optional.NewString(cmd.Branch),
-	}
-	keys, _, err := client.KeysApi.KeysList(Auth, cmd.ProjectID, &keysListLocalVarOptionals)
-	if err != nil {
-		return err
+	commonUnmentionedKeys := map[string]phrase.TranslationKey{}
+	alreadyInitialized := false
+	for _, id := range cmd.IDs {
+		q := "unmentioned_in_upload:" + id
```
theSoenke, line 37: not related to the PR but noticed that this does not seem to return anything when an upload itself didn't change anything

theSoenke (commented):
@jablan I've tried

clients/cli/cmd/internal/upload_cleanup.go
jablan: interesting. should we open a ticket? would that be TSS?

theSoenke: Probably it would even for integrations as we own the API. I'll open a ticket to take a closer look

theSoenke (approved):
Looks good!

PR #570, opened 2024-04-15
feat: Add linked-parent to translation details

BlumMichael:
Add "linked_parent" to translation-details.
https://phrase.atlassian.net/browse/TSS-3171


clients/cli/go.sum
malvarezphrase, line 1: **Comment**: Not sure about the changes in this file, but the rest LGTM!

malvarezphrase (approved):

clients/cli/go.sum
BlumMichael: That is actually a good question.
Can you help here please @theSoenke ?
I recompiled everything with `npm start` and committed every change. Is it okay to also add the changes in this file `clients/cli/go.sum`?  Why are those lines removed?

jablan: I don't know why are these lines removed. I don't think it would break anything if committed, but I'd keep it out of the PR.

malvarezphrase (approved):

PR #571, opened 2024-04-16
refactor!: add missing required params

theSoenke:
Improve documentation of required params. This will be a **breaking** change for some endpoints in the libraries and will be released in a new major version. It's **not** a breaking change for the REST API itself as these param were already required before and returned a 422 in case they were missing

The following endpoints now have additional required parameters:
`authorizations#create` - `note` is required
`blacklisted_keys#create` - `name` is required
`branches#create` - `name` is required
`distributions#create` - `name`, `project_id` are required
`figma_attachements#create` - `url` is required
`glossaries#create` - `name` is required
`glossary_term_translations#create` - `locale_code`, `content` are required
`glossary_terms#create` - `term` is required
`invitations#create` - `email`, `role` are required
`jobs#create` - `name` is required
`keys#create` - `name` is required
`locales#create` - `name`, `code` are required
`orders#create` - `name`, `lsp` are required
`projects#create` - `name` is required
`screenshot_markers#create` - `key_id` is required
`spaces#create` - `name` is required
`styleguides#create` - `title` is required
`tags#create` - `name` is required
`teams#add_project` - `id` is required
`team#add_space` - `id` is required
`teams#add_user` - `id` is required
`teams#create` - `name` is required
`translations#create` - `locale_id`, `key_id` are required
`uploads#create` - `file`, `file_format`, `locale_id` are required
`variables#create` - `name` is required
`webhooks#create` - `callback_url` is required

### DIFF
--- a/openapi-generator/templates/go/README.mustache
+++ b/openapi-generator/templates/go/README.mustache
@@ -17,7 +17,7 @@ import (
 	"context"
 	"fmt"
 
-	phrase "github.com/phrase/phrase-go/v3"
+	phrase "github.com/phrase/phrase-go/v2" // x-release-please-major
 )
 
 func main() {
--- a/openapi-generator/templates/go/README.mustache
+++ b/openapi-generator/templates/go/README.mustache
@@ -17,7 +17,7 @@ import (
 	"context"
 	"fmt"
 
-	phrase "github.com/phrase/phrase-go/v2"
+	phrase "github.com/phrase/phrase-go/v3"
 )
 
 func main() {
--- a/paths/distributions/create.yaml
+++ b/paths/distributions/create.yaml
@@ -48,6 +48,8 @@ requestBody:
       schema:
         type: object
         title: distribution/create/parameters
+        required:
+          - name
         properties:
           name:
             description: Name of the distribution
--- a/paths/distributions/create.yaml
+++ b/paths/distributions/create.yaml
@@ -50,6 +50,7 @@ requestBody:
         title: distribution/create/parameters
         required:
           - name
+          - project_id
         properties:
           name:
             description: Name of the distribution
--- a/paths/key_links/batch_destroy.yaml
+++ b/paths/key_links/batch_destroy.yaml
@@ -7,12 +7,6 @@ parameters:
 - "$ref": "../../parameters.yaml#/X-PhraseApp-OTP"
 - "$ref": "../../parameters.yaml#/project_id"
 - "$ref": "../../parameters.yaml#/key_id_as_id"
-- in: query
-  name: unlink_parent
-  required: false
-  description: Whether to unlink the parent key as well and unmark it as linked-key.
-  schema:
-    type: boolean
 requestBody:
   required: true
   content:
@@ -29,7 +23,10 @@ requestBody:
             example: ["child_key_id1", "child_key_id2"]
             items:
               type: string
-
+          unlink_parent:
+            description: Whether to unlink the parent key as well and unmark it as linked-key.
+            type: boolean
+            default: false
 responses:
   '200':
     description: OK
--- a/paths/orders/create.yaml
+++ b/paths/orders/create.yaml
@@ -48,6 +48,9 @@ requestBody:
       schema:
         type: object
         title: order/create/parameters
+        required:
+          - name
+          - lsp
         properties:
           branch:
             description: specify the branch to use
--- a/paths/orders/create.yaml
+++ b/paths/orders/create.yaml
@@ -65,11 +65,11 @@ requestBody:
             type: string
             example: textmaster
           source_locale_id:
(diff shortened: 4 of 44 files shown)


paths/comment_reactions/destroy.yaml
```diff
@@ -10,7 +10,7 @@ parameters:
 - "$ref": "../../parameters.yaml#/key_id"
 - "$ref": "../../parameters.yaml#/comment_id"
 - "$ref": "../../parameters.yaml#/id"
-- description: specify the branch to use
+- description: Branch name
```
jablan, line 13: perhaps this can be extracted to `parameters.yaml` and reused elsewhere?

paths/figma_attachment_keys/destroy.yaml
```diff
@@ -40,4 +34,16 @@ x-code-samples:
       --id <id> \
       --branch my-feature-branch \
```
jablan, line 35: this should then go to `--data` parameter instead, I guess?

paths/comment_reactions/destroy.yaml
theSoenke: Yep, lots of work but probably a good idea

paths/figma_attachment_keys/destroy.yaml
theSoenke: Yeah right. That makes me now more hesitant to move them all form query params to the body. Technically this is probably correct but also requires many changes and saw that I've still missed lots of branch query params

theSoenke: I've reverted these changes and just focused on required params

paths/distributions/create.yaml
```diff
@@ -48,6 +48,8 @@ requestBody:
       schema:
         type: object
         title: distribution/create/parameters
+        required:
+          - name
```
hahmed-dev, line 52: Was looking into `api/v2/distributions` [controller](https://github.com/phrase/phrase/blob/master/app/controllers/api/v2/distributions_controller.rb#L126), maybe I missed something, but it seems `project_id` if not provided, it won't set the `project`?
If thats the case, maybe `project_id` should be required as well?

theSoenke: good catch! that is required as well

paths/orders/create.yaml
```diff
@@ -48,6 +48,9 @@ requestBody:
       schema:
         type: object
         title: order/create/parameters
+        required:
+          - name
+          - lsp
```
hahmed-dev, line 53: `source_locale_id` should also be considered as mandatory? Considering its validated in the model?

paths/translations/create.yaml
```diff
@@ -54,7 +54,7 @@ requestBody:
             type: string
             example: my-feature-branch
           locale_id:
-            description: Locale. Can be the name or public id of the locale. Preferred is the public id.
+            description: Locale. Can be the name or id of the locale. Preferred is the id
```
hahmed-dev, line 57: Just a nit (absolutely not something that would block the merging).
Just think article `the` is not needed here.

```suggestion
            description: Locale. Can be the name or id of the locale. Preferred is id
```

openapi-generator/templates/go/README.mustache
```diff
@@ -17,7 +17,7 @@ import (
 	"context"
 	"fmt"
 
-	phrase "github.com/phrase/phrase-go/v2"
```
jablan, line 20: I believe this and the next change can be automatized using https://github.com/googleapis/release-please/blob/main/docs/customizing.md#updating-arbitrary-files

paths/key_links/batch_destroy.yaml
```diff
@@ -29,7 +23,10 @@ requestBody:
             example: ["child_key_id1", "child_key_id2"]
             items:
               type: string
-
+          unlink_parent:
+            description: Whether to unlink the parent key as well and unmark it as linked-key.
+            type: boolean
+            default: false
```
hahmed-dev, line 29: Just curious around the change to `default: false` from `required: false`. The way i see it, its always sending an extra args with default value. Is it because we are using `Boolifier` in the API?

openapi-generator/templates/go/README.mustache
theSoenke: that's really neat. trying it

paths/key_links/batch_destroy.yaml
theSoenke: the `default` param should be describing the API behaviour. If it's not passed the API will fallback to the default value

paths/orders/create.yaml
theSoenke: Just checked and if it's empty the project's default locale will be used

hahmed-dev (approved):

PR #578, opened 2024-04-17
feat(API): Add 'update_translation_keys' for Uploads [TSI-2292]

hahmed-dev:
Documents the newly introduced option `update_translation_keys` for `POST#Upload`.

**Related Issues:**
- https://phrase.atlassian.net/browse/TSI-2292
- https://phrase.atlassian.net/browse/TSI-2270


paths/uploads/create.yaml
```diff
@@ -85,6 +85,11 @@ requestBody:
             description: Indicates whether existing translations should be updated with the file content.
             type: boolean
             example:
+          update_translation_keys:
+            description: Indicates whether any missing translation keys should be created or existing ones updated during upload.
```
jablan, line 89: I'd rather phrase it as "pass `false` here to prevent new keys from being created and existing keys updated" or similar.

jablan (approved):

clients/java/src/test/java/com/phrase/client/api/UploadsApiTest.java
```diff
@@ -106,7 +107,7 @@ public void uploadCreateTest() throws ApiException, IOException, InterruptedExce
         Boolean autotranslate = null;
         Boolean markReviewed = null;
         Boolean tagOnlyAffectedKeys = null;
-        Upload response = api.uploadCreate(projectId, xPhraseAppOTP, branch, file, fileFormat, localeId, tags, updateTranslations, updateDescriptions, convertEmoji, skipUploadTags, skipUnverification, fileEncoding, localeMapping, formatOptions, autotranslate, markReviewed, tagOnlyAffectedKeys);
+        Upload response = api.uploadCreate(projectId, xPhraseAppOTP, branch, file, fileFormat, localeId, tags, updateTranslations, updateTranslationKeys, updateDescriptions, convertEmoji, skipUploadTags, skipUnverification, fileEncoding, localeMapping, formatOptions, autotranslate, markReviewed, tagOnlyAffectedKeys);
```
jablan, line 110: it's interesting that this was needed. I thought this parameter is not required? any idea why it's part of the signature?

hahmed-dev: I checked the newly built Java client, in fact other params that are marked as optional are all passed to the method. I guess all these arguments have to be passed to the method, we can set `null` for the ones that are optional.

hahmed-dev: Because if I am not wrong, a Java method would expect all the options to be passed, its within the method where we check if the option is present, then assign it to the payload.
Screenshot 2024-04-17 at 15 34 55

jablan: interesting, technically that means that any such change should be marked as breaking for java lib :( but I guess we won't be doing that.

jablan: I don't think so. This change shouldn't be a breaking one, we added an optional field. However, seems to me that Java generator is not correct so it generates a breaking change for Java. I don't think we have a way right now to declare breaking changes only for a single platform.
So, I'd keep this a minor version bump as usual, but we should be aware of this Java generator quirk, and perhaps try to address it in the future (perhaps updating the Java templates to newer version would be enough, although even that is no easy job).

infotexture:
@hahmed-dev How does this new parameter differ from `update_translations`? 🤔 

📖 It doesn't seem to be mentioned yet in the [CLI docs](https://support.phrase.com/hc/en-us/articles/5808300599068-Using-the-CLI-Strings#push-and-pull-0-6) or [sample configuration file](https://support.phrase.com/hc/article_attachments/13625178757788).

PR #534, opened 2024-02-01
fix(API): allow nullable value for job's due_date

hahmed-dev:
Purpose:
When creating a job from CLI and not providing `due_date` was adding a dummy due_date, which in fact was Go Lang adding a [zero value](https://go.dev/play/p/vTidOlmb9P) to an attribute `due_date` of Time type if no value is provided. 
This PR updates the schema to allow `nullable` values against `due_date` for `create`, `update` requests as well as when fetching the job.

Relevant Ticket: https://phrase.atlassian.net/browse/TSI-2297

PR #593, opened 2024-04-23
fix(Java): fix maven build

hahmed-dev:
In an earlier PR #534 to allow nullable values, a dependency in Java client was introduced, ever since it being merged, build on `java-client` repo has been failing complaining about missing dependency.

Possible reason can be, we use `gradle` within OpenApi whereas `java-client` uses `maven` and the said dependency had to be included in the `pom.xml` file as well. For now have added the dependency to `pom.mustache` with the latest release version.

[Link to failed builds on java-client](https://github.com/phrase/phrase-java/actions/runs/7781883827/job/21217194464#step:4:136)

### DIFF
--- a/openapi-generator/templates/java/pom.mustache
+++ b/openapi-generator/templates/java/pom.mustache
@@ -363,6 +363,11 @@
             <version>${junit-version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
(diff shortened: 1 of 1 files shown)


theSoenke (approved):
nice thanks!

jablan (approved):

PR #616, opened 2024-05-14
chore: install latest phrase-go with go get

theSoenke:
Update readme to list phrase-go v3 https://github.com/phrase/phrase-go/pull/11


jablan (approved):
would it make sense to add release-please magical comment here too? we could probably use mustache comments so that it doesn't end up in the resulting README at all.

theSoenke:
@jablan tried adding the magical comments. And thanks for the mustache comment hint. Used it incorrectly before and fixed it

jablan (approved):
:magic_wand:

PR #618, opened 2024-05-31
fix(CLI): use id instead of repo_sync_id as param

theSoenke:
The param to work with a resource should be called simply `id`


jablan (approved):
thanks

PR #622, opened 2024-06-10
feat(API): Add OTA Release Triggers API [TSI-2485]

jablan:
https://github.com/phrase/phrase/pull/13499

### DIFF
--- a/paths/release_triggers/create.yaml
+++ b/paths/release_triggers/create.yaml
@@ -26,7 +26,7 @@ requestBody:
             type: string
             example: "Europe/Berlin"
           locale_ids:
-            description: List of locale ids that will be included in the release. If empty, distribution locales will be used
+            description: List of locale ids that will be included in the release.
             type: array
             items:
               type: string
--- a/paths/release_triggers/create.yaml
+++ b/paths/release_triggers/create.yaml
@@ -1,0 +1,91 @@
+---
+summary: Create a release trigger
+description: Create a new recurring release. New releases will be published automatically, based on the cron schedule provided.
+  Currently, only one release trigger can exist per distribution.
+operationId: release_triggers/create
+tags:
+  - Release Triggers
+parameters:
+  - "$ref": "../../parameters.yaml#/X-PhraseApp-OTP"
+  - "$ref": "../../parameters.yaml#/account_id"
+  - "$ref": "../../parameters.yaml#/distribution_id"
+requestBody:
+  required: true
+  content:
+    application/json:
+      schema:
+        type: object
+        title: release/create/parameters
+        properties:
+          cron_schedule:
+            description: Cron schedule for the scheduler. Read more about the format of this field at https://en.wikipedia.org/wiki/Cron
+            type: string
+            example: "15 18 1,3 * *"
+          time_zone:
+            description: Time zone for the scheduler
+            type: string
+            example: "Europe/Berlin"
+          locale_ids:
+            description: List of locale ids that will be included in the release. If empty, distribution locales will be used
+            type: array
+            items:
+              type: string
+            example:
+              - abcd1234cdef1234abcd1234cdef1234
+              - fff565db236400772368235db2c6117e
+          tags:
+            description: Only include tagged keys in the release. For a key to be included it must be tagged with all tags provided
+            type: array
+            items:
+              type: string
+            example:
+              - android
+              - feature1
+          branch:
+            description: Branch used for release
+            type: string
+            example: my-feature-branch
+          app_min_version:
+            description: The created releases will be available only for apps with version greater or equal to this value
+            type: string
+            example: "1.0.0"
+          app_max_version:
+            description: The created releases will be available only for apps with version less or equal to this value
+            type: string
+            example: "2.0.0"
+responses:
+  "201":
+    description: Created
+    content:
+      application/json:
+        schema:
+          "$ref": "../../schemas/release_trigger.yaml#/release_trigger"
+    headers:
+      X-Rate-Limit-Limit:
+        "$ref": "../../headers.yaml#/X-Rate-Limit-Limit"
+      X-Rate-Limit-Remaining:
+        "$ref": "../../headers.yaml#/X-Rate-Limit-Remaining"
+      X-Rate-Limit-Reset:
+        "$ref": "../../headers.yaml#/X-Rate-Limit-Reset"
+  "400":
+    "$ref": "../../responses.yaml#/400"
+  "404":
+    "$ref": "../../responses.yaml#/404"
+  "429":
+    "$ref": "../../responses.yaml#/429"
+x-code-samples:
+  - lang: Curl
+    source: |-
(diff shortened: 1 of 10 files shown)


paths/release_triggers/create.yaml
```diff
@@ -0,0 +1,91 @@
...
+          cron_schedule:
+            description: Cron schedule for the scheduler. Read more about the format of this field at https://en.wikipedia.org/wiki/Cron
+            type: string
+            example: "15 18 1,3 * *"
+          time_zone:
+            description: Time zone for the scheduler
+            type: string
+            example: "Europe/Berlin"
+          locale_ids:
+            description: List of locale ids that will be included in the release. If empty, distribution locales will be used
```
hahmed-dev, line 29: Actually thats not the case atm, we are not falling back to distribution's locales if the params are nil. Should we? 🤔

paths/release_triggers/index.yaml
```diff
@@ -0,0 +1,52 @@
...
+  Note: Currently only one release trigger can exist per distribution.
+operationId: release_triggers/list
+tags:
+- Release Triggers
+parameters:
+- "$ref": "../../parameters.yaml#/X-PhraseApp-OTP"
+- "$ref": "../../parameters.yaml#/account_id"
+- "$ref": "../../parameters.yaml#/distribution_id"
+- "$ref": "../../parameters.yaml#/page"
+- "$ref": "../../parameters.yaml#/per_page"
```
hahmed-dev, line 14: Not sure if we are paginating the response in API.

paths/release_triggers/show.yaml
```diff
@@ -0,0 +1,44 @@
+---
+summary: Get a single release trigger
+description: Get details on a single release trigger.
```
hahmed-dev, line 3: small typo I believe. 
```suggestion
description: Get details of a single release trigger.
```

paths/release_triggers/create.yaml
jablan: I'll remove that detail for now, I believe it should fall back, but again, a product decision I guess.

paths/release_triggers/create.yaml
```diff
@@ -0,0 +1,91 @@
...
+  content:
+    application/json:
+      schema:
+        type: object
+        title: release/create/parameters
+        properties:
+          cron_schedule:
+            description: Cron schedule for the scheduler. Read more about the format of this field at https://en.wikipedia.org/wiki/Cron
+            type: string
+            example: "15 18 1,3 * *"
```
hahmed-dev, line 23: Not something that should block this PR, but since we don't offer a variety of scheduling options specifically around days of months/months, should our examples also reflect it in our example? In our case, that would mean the third and fourth entry of cron schedule will always be "*"

e.g. 00 11 * * 0,4,5,6

paths/release_triggers/create.yaml
```diff
@@ -0,0 +1,91 @@
...
+    "$ref": "../../responses.yaml#/404"
+  "429":
+    "$ref": "../../responses.yaml#/429"
+x-code-samples:
+  - lang: Curl
+    source: |-
+      curl "https://api.phrase.com/v2/accounts/:account_id/distributions/:distribution_id/release_triggers" \
+        -u USERNAME_OR_ACCESS_TOKEN \
+        -X POST \
+        -d '{"cron_schedule":"15 18 1,3 * *","time_zone":"Europe/Berlin","locale_ids":["abcd1234cdef1234abcd1234cdef1234","fff565db236400772368235db2c6117e"],"tags":["android","feature1"],"branch":"my-feature-branch","app_min_version":"1.0.0","app_max_version":"2.0.0"}' \
```
hahmed-dev, line 82: same as above if we want to change the cron example to align more with our use cases:

```suggestion
        -d '{"cron_schedule":"15 18 * * 0,1","time_zone":"Europe/Berlin","locale_ids":["abcd1234cdef1234abcd1234cdef1234","fff565db236400772368235db2c6117e"],"tags":["android","feature1"],"branch":"my-feature-branch","app_min_version":"1.0.0","app_max_version":"2.0.0"}' \
```

paths/release_triggers/create.yaml
hahmed-dev: Well on a second thought, I guess this shouldn't be an issue, allowing user a flexibility via the API, this should be translated into appropriate next_run_at by Fugit without issues and release will be triggered accordingly ideally.. So I guess if anything this is more of a flexibility that is on API that is not present on UI. What do you think?

paths/release_triggers/create.yaml
```diff
@@ -0,0 +1,91 @@
...
+            type: string
+            example: my-feature-branch
+          app_min_version:
+            description: The created releases will be available only for apps with version greater or equal to this value
+            type: string
+            example: "1.0.0"
+          app_max_version:
+            description: The created releases will be available only for apps with version less or equal to this value
+            type: string
+            example: "2.0.0"
```
hahmed-dev, line 55: I apologise for going back and forth on this PR, I think `platform` should also be documented among other params?

paths/release_triggers/update.yaml
```diff
@@ -0,0 +1,92 @@
...
+          cron_schedule:
+            description: Cron schedule for the scheduler. Read more about the format of this field at https://en.wikipedia.org/wiki/Cron
+            type: string
+            example: "15 18 1,3 * *"
+          time_zone:
+            description: Time zone for the scheduler
+            type: string
+            example: "Europe/Berlin"
+          locale_ids:
+            description: List of locale ids that will be included in the release. If empty, distribution locales will be used
```
hahmed-dev, line 29: I think we agreed for now to remove fallback locale_ids until we have some further clarity on it

```suggestion
            description: List of locale ids that will be included in the release.
```

hahmed-dev (commented):
Just final two comments!

paths/release_triggers/create.yaml
jablan: hm IMO we should offer the same features through UI and API, so e.g. not allow triggering stuff every minute etc.

not sure how easy this would be to validate (and reject the request if needed), perhaps fugit offers something...

paths/release_triggers/create.yaml
jablan: hm I don't see where do we assign it upon creation in the `NestedAttributesBuilder`?

hahmed-dev: Ah, we just use the distribution's platforms to fetch the formats. Normal release form has a checkboxes to check which one, this release will be for.

paths/release_triggers/create.yaml
hahmed-dev: Like having a validation to raise error when cron has for example entries for month/days of month other than *?

jablan: yeah, or multiple minutes/hours given etc

jablan: I adapted the example already, I think

hahmed-dev: yes you did! my bad missed it.

PR #632, opened 2024-06-12
build(go): Update goreleaser config file

jablan:


theSoenke (approved):
@jablan maybe we can also lock the version to update on our own time in the future? https://github.com/phrase/phrase-go/blob/master/.github/workflows/release.yml#L22

jablan:
:see_no_evil: I missed this part, I would have just pinned it to 1.x and skipped the whole upgrade.

clients/go/.github/workflows/release.yml
```diff
@@ -17,9 +17,9 @@ jobs:
         with:
           go-version: 1.14
       - name: Run GoReleaser
-        uses: goreleaser/goreleaser-action@ac067437f516133269923265894e77920c3dce18 # pin@v2
+        uses: goreleaser/goreleaser-action@v6
```
theSoenke, line 20: We should still pin it to a commit for security purposes

theSoenke (approved):

PR #649, opened 2024-06-25
feat(CLI): Add option for async download [TSI-2515]

jablan:
If option `--async` is passed with `pull` command, we will use async locale downloads, in order to handle large exports which normally cause timeouts on API.

The async download is initiated, then the API is queried every 5 seconds until the processing is done and the file is ready to be downloaded.

Then we download the file (using standard go's HTTP client, but passing the token as it's currently also performed through the API, which might be changed in the future).

Ticket: https://phrase.atlassian.net/browse/TSI-2515


clients/cli/cmd/internal/pull.go
```diff
@@ -169,33 +173,128 @@ func (target *Target) DownloadAndWriteToFile(client *phrase.APIClient, localeFil
...
+
+		localVarOptionals := phrase.LocaleDownloadCreateOpts{}
+		if Debug {
+			fmt.Fprintln(os.Stderr, "Initiating async download...")
+		}
+		asyncDownload, _, _ := client.LocaleDownloadsApi.LocaleDownloadCreate(Auth, target.ProjectID, localeFile.ID, localeDownloadCreateParams, &localVarOptionals)
+
+		for asyncDownload.Status == "processing" {
+			if Debug {
+				fmt.Fprintln(os.Stderr, "Waiting for the files to be exported...")
```
hahmed-dev, line 187: Was thinking if we can have some sort of function roughly like

```
func debugLog(debugMessage string) {
  if Debug {
    fmt.Fprintln(os.Stderr, debugMessage)
  }
}
```

And use it in place of conditionally logging debug messages? I understand this is existing way of logging debug messages but I was just thinking as a starting point this can to some extent do code clean up.

hahmed-dev: Or we can have it as a future improvement, where this Debug function can be exposed by some other package or same file within the same package and used app wide.

clients/cli/cmd/internal/pull.go
```diff
@@ -169,33 +173,128 @@ func (target *Target) DownloadAndWriteToFile(client *phrase.APIClient, localeFil
...
+
+func asyncDownloadParams(localVarOptionals phrase.LocaleDownloadOpts) phrase.LocaleDownloadCreateParameters {
+	localeDownloadCreateParams := phrase.LocaleDownloadCreateParameters{}
+
+	localeDownloadCreateParams.Branch = localVarOptionals.Branch.Value()
+	localeDownloadCreateParams.FileFormat = localVarOptionals.FileFormat.Value()
+	localeDownloadCreateParams.Tags = localVarOptionals.Tags.Value()
+	if localVarOptionals.IncludeEmptyTranslations.IsSet() {
+		localeDownloadCreateParams.IncludeEmptyTranslations = new(bool)
+		*localeDownloadCreateParams.IncludeEmptyTranslations = localVarOptionals.IncludeEmptyTranslations.Value()
```
hahmed-dev, line 261: Question here (mostly for my personal knowledge), since these attributes on type e.g. `IncludeEmptyTranslations` already have a datatype assigned to it in the struct defintion, is it still required to initialize them with `new(bool)`?

hahmed-dev (commented):
For the most part 👌 , just a few comments, also thinking, is it worth creating sub functions for sync/async download and call those conditionally inside `DownloadAndWriteToFile`?

clients/cli/cmd/internal/pull.go
jablan: I think yes, as it's defined as a bool pointer.

clients/cli/cmd/internal/pull.go
```diff
@@ -155,47 +157,146 @@ func (target *Target) DownloadAndWriteToFile(client *phrase.APIClient, localeFil
...
+	localeDownloadCreateParams := asyncDownloadParams(downloadOpts)
+
+	localVarOptionals := phrase.LocaleDownloadCreateOpts{}
+	debugFprintln("Initiating async download...")
+	asyncDownload, _, err := client.LocaleDownloadsApi.LocaleDownloadCreate(Auth, target.ProjectID, localeFile.ID, localeDownloadCreateParams, &localVarOptionals)
+	if err != nil {
+		return err
+	}
+
+	for asyncDownload.Status == "processing" {
```
theSoenke, line 189: Thinking whether also should introduce a total timeout 🤔 Like 30 minutes for example

clients/cli/cmd/internal/pull.go
```diff
@@ -155,47 +157,146 @@ func (target *Target) DownloadAndWriteToFile(client *phrase.APIClient, localeFil
...
+	response, err := http.DefaultClient.Do(request)
+	if err != nil {
+		return err
+	}
+	defer response.Body.Close()
+	io.Copy(file, response.Body)
+	return nil
+}
+
+func asyncDownloadParams(localVarOptionals phrase.LocaleDownloadOpts) phrase.LocaleDownloadCreateParameters {
```
theSoenke, line 255: I think we have to somehow automatically convert `phrase.LocaleDownloadOpts` to `phrase.LocaleDownloadCreateParameters`. Otherwise we will always play catch-up when a new download option is introduced and will have to add it here

clients/cli/cmd/internal/pull.go
```diff
@@ -155,47 +157,146 @@ func (target *Target) DownloadAndWriteToFile(client *phrase.APIClient, localeFil
...
+	localeDownloadCreateParams.FileFormat = localVarOptionals.FileFormat.Value()
+	localeDownloadCreateParams.Tags = localVarOptionals.Tags.Value()
+	if localVarOptionals.IncludeEmptyTranslations.IsSet() {
+		localeDownloadCreateParams.IncludeEmptyTranslations = new(bool)
+		*localeDownloadCreateParams.IncludeEmptyTranslations = localVarOptionals.IncludeEmptyTranslations.Value()
+	}
+	if localVarOptionals.ExcludeEmptyZeroForms.IsSet() {
+		localeDownloadCreateParams.ExcludeEmptyZeroForms = new(bool)
+		*localeDownloadCreateParams.ExcludeEmptyZeroForms = localVarOptionals.ExcludeEmptyZeroForms.Value()
+	}
```
hahmed-dev, line 268: Last point from my side, if we can create setters for setting boolean and json attributes, I think we can further cleanup this function.

clients/cli/cmd/internal/pull.go
jablan: with a great help of Copilot :pilot: :sweat_smile: , I managed to rewrite it to use reflection. I tried some basic cases and it seems to work.

theSoenke (approved):

PR #657, opened 2024-07-24
build(go, CLI): Add license check for go and CLI [TSI-2537]

jablan:


.github/workflows/test-cli.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Generate CLI
+        run: |
+          npm install
+          npm run generate.cli
+      - name: License check
+        uses: phrase/actions/lawa-ci
+        with:
+          github-token: ${{ secrets.GH_LAWA_ACCESS_TOKEN }}
+          decisions-file: .license-decisions.yml
+          ruby-version: 3.3.2
```
theSoenke, line 24: Should we set this for non-Ruby projects? Maybe a default would be better here?

jablan: you're right, no need to override this

theSoenke (approved):

PR #661, opened 2024-07-29
fix(go): Fix query param serialization of nested objects [TSI-2594]

bikmazefe:
Fix serializing format_options when downloading locales. We're using a nested structure for `custom_metadata_columns` which were being wrongfully sent as `{"custom_metadata_columns"=>"map[foo:5]"}`.

Similar to the issue we fixed for uploads in https://github.com/phrase/openapi/pull/498

https://phrase.atlassian.net/browse/TSI-2594


jablan (approved):
nice!

theSoenke (approved):

PR #669, opened 2024-08-15
chore: warn users about sending empty filter values when deleting keys

hahmed-dev:
Slack [thread](https://memsource.slack.com/archives/C04H8PB09UN/p1723710962243209) for more context.

Screenshot 2024-08-15 at 13 41 32


theSoenke:
@hahmed-dev thanks for picking this up. But i'm wondering whether adding it for the key name is enough. The same is the case for all other parameters. If the tags are empty also no tag filtering will be applied. Same for upload_id etc. Maybe we can add a general sentence below the params?

hahmed-dev:
@theSoenke Good point, updated.

Screenshot 2024-08-15 at 15 21 18

paths/keys/destroy-collection.yaml
```diff
@@ -24,6 +24,9 @@ parameters:
       <li><code>updated_at:{>=|<=}2013-02-21T00:00:00Z</code> for date range queries</li>
       <li><code>unmentioned_in_upload:upload_id</code> to filter keys unmentioned within upload</li>
     </ul>
+    <br/>
+    <p><strong>Caution:</strong> If an empty value is sent for any of these filters, all records will be deleted as it is treated as selecting all records. Please ensure a specific value is provided to avoid accidental data loss.</p>
```
theSoenke, line 28: Or how about? Otherwise it might sound you have to provide a value for every single filter 😅 
> Parameters with empty values will be ignored and have the same effect as if the parameter is not present

hahmed-dev: True.

theSoenke (approved):

PR #670, opened 2024-08-26
feat: Add update_translations_on_source_match

ildarsafin:
Add `update_translations_on_source_match` to update target translations only if the source translations match the uploaded multilingual file.

### DIFF
--- a/paths/uploads/create.yaml
+++ b/paths/uploads/create.yaml
@@ -95,7 +95,7 @@ requestBody:
             default: true
             example:
           update_translations_on_source_match:
-            description: Pass `true` to update target translations only if the source translations match the uploaded multilingual file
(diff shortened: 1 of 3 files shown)


paths/uploads/create.yaml
```diff
@@ -94,6 +94,11 @@ requestBody:
             type: boolean
             default: true
             example:
+          update_translations_on_source_match:
+            description: Pass `true` to update target translations only if the source translations match the uploaded multilingual file
```
forelabs, line 98: I would leave out the "pass `true` wording, as we don't use it anywhre else and for boolean flags this should be clear already.

```suggestion
            description: Update target translations only if the source translations of the uploaded multilingual file match the stored translations
```

Also I think the sentence can include a bit more details for the match logic

forelabs (approved):

PR #678, opened 2024-08-26
feat: Document locale_ids param

forelabs:
- Related to https://phrase.atlassian.net/browse/TSS-3542

Required for orchestrator to pick up the new parameter

### DIFF
--- a/paths/locale_downloads/create.yaml
+++ b/paths/locale_downloads/create.yaml
@@ -1,4 +1,3 @@
----
 summary: Initiate async download of a locale
 description: Prepare a locale for download in a specific file format.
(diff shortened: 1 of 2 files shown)


paths/locale_downloads/create.yaml
```diff
@@ -101,7 +108,6 @@ responses:
 x-code-samples:
   - lang: Curl
     source: |-
-
```
forelabs, line 104: sorry got removed with auto formatting, not sure if I should remove it or leave it. I think having the newline as code example should not be needed

hahmed-dev (approved):

PR #683, opened 2024-09-09
build(CLI): Run test workflow only for CLI releases

bikmazefe:
Run the CLI tests only when there are related changes, otherwise it fails due to missing Go package version.

### DIFF
--- a/.github/workflows/test-cli.yml
+++ b/.github/workflows/test-cli.yml
@@ -1,7 +1,10 @@
 # Currently this does not contain any tests, only a license check
 # Once we add tests, we should add a step here to run them
 name: Run CLI Tests
-on: [push]
+on:
+  push:
+    paths:
+      - openapi-generator/cli_lang.yaml
 jobs:
   test:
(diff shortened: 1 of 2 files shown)


theSoenke (approved):
🙌

jablan:
hm what about changes to CLI code itself?

theSoenke:
@jablan @bikmazefe maybe we could then add `clients/cli/**` to the paths?

bikmazefe:
@jablan @theSoenke  Ah good point, I was thinking that the changes in Go would cause the version bump in the yaml file anyway but there are some static code there as well. Thank you, I'll add it.

bikmazefe:
It seems like the latest version of the `tools` module that was released on September 9th is causing [some issues](https://github.com/phrase/openapi/actions/runs/10832138504/job/30055743782#step:4:818) in the CI. I fixed it's version to the previous one. Can you please have another look? 🙏 @theSoenke @jablan

theSoenke (approved):

PR #688, opened 2024-09-25
fix(java): Properly encode nested parameters with file upload

jablan:
Java: nested parameters (e.g. `format_options` in "create upload" call) were not properly encoded when constructing multipart request.

https://phrase.atlassian.net/browse/STRINGS-411


theSoenke (approved):
Looks good ✅  Wondering whether we can also improve the generated docs to use HashMap instead of Object?

PR #698, opened 2024-10-02
chore: disable golang proxy for phrase-go

theSoenke:
Requests are cached for 30 minutes when a non existing version is requested. This is causing lots of build failures during the time when a new phrase-go version is published and used. By setting `GOPRIVATE` the proxy should be skipped and the package directly fetched from GitHub


jablan (approved):
interesting, nice find.

PR #701, opened 2024-10-18
feat(CLI): Support branch in pull config #STRINGS-538

jablan:
Supports getting the branch from pull target configuration.

https://github.com/phrase/phrase-cli/issues/152

### DIFF
--- a/clients/cli/cmd/internal/push_source.go
+++ b/clients/cli/cmd/internal/push_source.go
@@ -148,10 +148,10 @@ func (source *Source) CheckPreconditions() error {
 	return nil
 }
 
-func (sources Sources) ProjectIdsBranches() []LocaleCacheKey {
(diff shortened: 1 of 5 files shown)


clients/cli/cmd/internal/push_source.go
```diff
@@ -148,12 +148,12 @@ func (source *Source) CheckPreconditions() error {
 	return nil
 }
 
-func (sources Sources) ProjectIds() []string {
-	projectIds := []string{}
+func (sources Sources) ProjectIdsBranches() []LocaleCacheKey {
```
theSoenke, line 151: This function name confused me a bit because it returns a `LocaleCacheKey`. Maybe we could either rename the function or the struct name so that it's more clear what it returns?

theSoenke (approved):
Looks good!

PR #703, opened 2024-11-08
fix(PHP): Fix serializing nested form parameters #STRINGS-730

jablan:
Related to https://github.com/phrase/openapi/pull/130

This PR adds better support for serializing form parameters (e.g. when uploading file through API):
* passing in arrays and objects
* support nested structures


theSoenke (approved):
LGTM ✅

PR #718, opened 2024-11-20
fix(API): Comment creation schema fix #STRINGS-866

jablan:
https://phrase.atlassian.net/browse/STRINGS-866

### DIFF
--- a/doc/compiled.json
+++ b/doc/compiled.json
@@ -13093,27 +13093,6 @@
           },
           {
             "$ref": "#/components/parameters/key_id"
-          },
-          {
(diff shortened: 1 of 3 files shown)


theSoenke:
@jablan i believe for this endpoint it's specifically an issue because it passes an array via query params. Are there by any chance more endpoints with similar issues? 😅 And does it have an impact on the client interfaces?

theSoenke:
@jablan thanks for checking. Yeah, then this have to be a breaking change

theSoenke (approved):

PR #722, opened 2024-11-25
fix(CLI): crash when target/params section is empty #STRINGS-921

jablan:
addresses https://github.com/phrase/phrase-cli/issues/157

### DIFF
--- a/clients/cli/cmd/internal/pull_target.go
+++ b/clients/cli/cmd/internal/pull_target.go
@@ -18,7 +18,7 @@ type Targets []*Target
 func (targets Targets) GetAllLocalesCacheKeys() []LocalesCacheKey {
 	localesCacheKeys := []LocalesCacheKey{}
 	for _, target := range targets {
-		localesCacheKeys = append(localesCacheKeys, LocalesCacheKey{target.ProjectID, target.Params.Branch.Value()})
+		localesCacheKeys = append(localesCacheKeys, LocalesCacheKey{target.ProjectID, target.GetBranch()})
 	}
 	return localesCacheKeys
 }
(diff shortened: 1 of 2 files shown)


clients/cli/cmd/internal/pull_target.go
```diff
@@ -32,6 +32,13 @@ type Target struct {
 	RemoteLocales []*phrase.Locale
 }
 
+func (target *Target) GetBranch() string {
+	if target.Params != nil {
+		return target.Params.Branch.Value()
+	}
+	return ""
```
theSoenke, line 39: Wondering whether this should be an empty string or nil when there is no branch? You could return nil when the return type is *string

jablan: that would be quite a bigger change, but I can try

jablan: @theSoenke sorry, I took a look, basically everywhere, `branch` is a string (not a pointer), so changing it would be a big effort

Things like:
https://github.com/phrase/openapi/blob/d9c8271a1d7440aaf600e46bc4d3a6c159651d5e/clients/cli/cmd/internal/pull.go#L33
https://github.com/phrase/openapi/blob/30ae809b38ead0d0d019362b67a8b604ac15fe5f/clients/cli/cmd/internal/push.go#L28
https://github.com/phrase/openapi/blob/30ae809b38ead0d0d019362b67a8b604ac15fe5f/clients/cli/cmd/internal/push_source.go#L199

etc etc

theSoenke: no worries! just wanted to check whether it makes sense

theSoenke (approved):

PR #733, opened 2024-12-12
feat(API): Add 'default_encoding' documentation

BlumMichael:
Documentation for https://github.com/phrase/phrase/pull/14383/

image

### DIFF
--- a/paths/projects/update.yaml
+++ b/paths/projects/update.yaml
@@ -141,4 +141,8 @@ requestBody:
             description: "(Optional) Requires autotranslate_enabled to be true"
             type: boolean
             example: true
+          default_encoding:
+            description: "(Optional) Sets the default encoding for Uploads. If you leave it empty, we will try to guess it automatically for you when you Upload a file. You can still overwrite this value by setting the <a href='#post-/projects/-project_id-/uploads'>`file_encoding`</a> parameter for Uploads."
+            type: boolean
+            example: true            
 x-cli-version: "2.6.3"
--- a/paths/projects/update.yaml
+++ b/paths/projects/update.yaml
@@ -143,6 +143,12 @@ requestBody:
             example: true
           default_encoding:
             description: "(Optional) Sets the default encoding for Uploads. If you leave it empty, we will try to guess it automatically for you when you Upload a file. You can still overwrite this value by setting the <a href='#post-/projects/-project_id-/uploads'>`file_encoding`</a> parameter for Uploads."
-            type: boolean
-            example: true            
+            type: string
+            example: "UTF-8"            
+            enum:
+              - "UTF-8"
+              - "UTF-16"
+              - "UTF-16BE"
+              - "UTF-16LE"
+              - "ISO-8859-1"
 x-cli-version: "2.6.3"
--- a/paths/projects/update.yaml
+++ b/paths/projects/update.yaml
@@ -142,7 +142,7 @@ requestBody:
             type: boolean
             example: true
           default_encoding:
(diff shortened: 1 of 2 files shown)


paths/projects/update.yaml
```diff
@@ -141,4 +141,8 @@ requestBody:
             description: "(Optional) Requires autotranslate_enabled to be true"
             type: boolean
             example: true
+          default_encoding:
+            description: "(Optional) Sets the default encoding for Uploads. If you leave it empty, we will try to guess it automatically for you when you Upload a file. You can still overwrite this value by setting the <a href='#post-/projects/-project_id-/uploads'>`file_encoding`</a> parameter for Uploads."
+            type: boolean
```
theSoenke, line 146: I believe the type should be string? And could you also mention valid examples?

theSoenke:
@BlumMichael is the param also exposed via the show project endpoint and should be documented here?

paths/projects/update.yaml
BlumMichael: Fixed the type, thank you

BlumMichael:
> @BlumMichael is the param also exposed via the show project endpoint and should be documented here?

It's not exposed via show-endpoint: https://github.com/phrase/phrase/blob/STRINGS-843-follow-up/app/serializers/project_serializer.rb

paths/projects/update.yaml
```diff
@@ -141,4 +141,14 @@ requestBody:
             description: "(Optional) Requires autotranslate_enabled to be true"
             type: boolean
             example: true
+          default_encoding:
+            description: "(Optional) Sets the default encoding for Uploads. If you leave it empty, we will try to guess it automatically for you when you Upload a file. You can still overwrite this value by setting the <a href='#post-/projects/-project_id-/uploads'>`file_encoding`</a> parameter for Uploads."
```
jablan, line 145: "override" instead of "overwrite" :)

theSoenke:
@BlumMichael i think we should then also add it to the project show endpoint? if users want to set it they probably also want to check the current state :)

jablan:
@theSoenke create too, I guess?

BlumMichael:
> @BlumMichael i think we should then also add it to the project show endpoint? if users want to set it they probably also want to check the current state :)

@theSoenke 

None of the other settings (e.g. `"smart_suggest_enabled"`, `"enable_branching"`, `"autotranslate_enabled"`, and so on) are available on the `show` endpoint. While I agree that it might be useful to have them there, I think it is out of scope for this ticket. Maybe lets create a separate task for that?

### Example show-call 
`http://localhost:3000/api/v2/projects/6033a78e0e993a15409ec9d18fa93bb2`
```json
{
    "id": "6033a78e0e993a15409ec9d18fa93bb2",
    "name": "Strings-563",
    "slug": "strings-563",
    "main_format": "strings_json",
    "media": "",
    "shares_translation_memory": true,
    "project_image_url": null,
    "created_at": "2024-10-29T06:29:08Z",
    "updated_at": "2024-12-12T10:08:11Z",
    "account": {
        "id": "c1d4a2bcaa4aa8b8a0ec9f5cf23cc322",
        "name": "Phrase2",
        "slug": "phrase",
        "company": "",
        "created_at": "2024-08-08T06:31:46Z",
        "updated_at": "2024-12-10T08:51:17Z",
        "company_logo_url": null
    },
    "branch": null,
    "space": null,
    "point_of_contact": {
        "id": "66d2f0e1b74143a05630e79c9f4b0912",
        "username": "michael.blum+1000",
        "name": "Michael Blum",
        "gravatar_uid": "8c6376c873fdf69071a979f8e4be01b5"
    }
}
```

theSoenke:
@BlumMichael sure, then let's make them all public in a follow-up task! ✅

theSoenke (approved):

BlumMichael:
> @BlumMichael sure, then let's make them all public in a follow-up task! ✅

:raised_hands: 
Added https://phrase.atlassian.net/browse/STRINGS-1029

PR #735, opened 2024-12-13
feat!: Remove old Git sync endpoints. Replaced with new repo sync

theSoenke:
BREAKING CHANGE: Remove old Git sync endpoints from the documentation. These have been replaced with the new repo sync: https://developers.phrase.com/api/#get-/accounts/-account_id-/repo_syncs Existing API calls will keep working but this is a breaking change for the SDKs as they will no longer support the old GitHub, GitLab and Bitbucket endpoints


jablan (approved):
:crossed_fingers: this also gives us the opportunity to introduce other breaking changes, if we have some pending?

theSoenke:
@jablan i'm not aware of something right now. at least nothing that wouldn't require large changes in the generator

PR #740, opened 2024-12-18
fix(go): Fix go module in tests

theSoenke:
Also bump go module version in tests


jablan (approved):
thanks, no idea why the magic comment doesn't work

PR #747, opened 2024-12-19
fix(API): Match repo_sync events to the correct response

horstmannmat:
We use `phrase-cli` in our CI/CD pipeline to automate merges to our repositories. However, the recent changes introduced in #735 have broken our process.  
While upgrading our pipeline to use `repo_sync`, I encountered the following issues with `phrase-cli`:  

### Issues Description 

---

#### **Missing `event_type` and `pull_request_url`**  
- The `phrase-cli` does not provide the `event_type` of the sync event or the `pull_request_url`. These details are critical for avoiding multiple merge requests in our CI/CD.  
- The API returns `type` instead of `event_type`, which caused discrepancies in the schemas. I’ve updated the schemas in this PR to match the API response.  

##### Reproduction  
Example with `phrase-cli`:  
```bash
$ phrase repo_syncs show --id $SYNC_EVENT_ID --repo_sync_id $GITLAB_SYNC_ID --account_id $ACCOUNT_ID
{
  "id": "<REDACTED>",
  "created_at": "2024-12-19T17:35:52Z",
  "status": "success",
  "auto_import": false
}
```

Example with curl (fetching type but missing pull_request_url):

```bash
 curl -X GET "https://api.phrase.com/v2/accounts/ $ACCOUNT_ID/repo_syncs/$GITLAB_SYNC_ID/events/$SYNC_EVENT_ID?access_token=$PHRASE_ACCESS_TOKEN"  -H "Accept: application/json"
{
     "id": "<REDACTED>",
     "type":"export",
     "created_at":"2024-12-19T17:35:52Z",
     "status":"success",
     "auto_import":false
}
```

I’ve updated the schemas in this PR to match the API response.

--- 

#### **Command Parsing conflicts:**:

- When parsing the commands in [api_repo_syncs.go#L169](https://github.com/phrase/phrase-cli/blob/master/cmd/api_repo_syncs.go#L169) and [api_repo_syncs.go#L499](https://github.com/phrase/phrase-cli/blob/master/cmd/api_repo_syncs.go#L499), a conflict arises as shown in (this snippet)[https://go.dev/play/p/DoeBHQOf-oI]

- For instance, the output of the help command demonstrates duplicate show commands:
```bash
> $ phrase repo_syncs --help                                                                                 ⬡ 20.12.1
RepoSyncs API

Usage:
  phrase repo_syncs [command]

Available Commands:
  activate    Activate a Repo Sync
  deactivate  Deactivate a Repo Sync
  events      Repository Syncs History
[...]

PR #748, opened 2024-12-20
fix(API): Repo Sync schema fixes #STRINGS-1074

jablan:
Hey @horstmannmat, I changed slightly your PR https://github.com/phrase/openapi/pull/747 by extracting "Repo Sync Events" to a separate section (IMO it's more RESTful that way and makes future extending of the events API a bit easier).

Would you mind telling me what you think, would that work for you?

cc @theSoenke

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,4 @@
 golang 1.21.9
 nodejs 20.11.0
 maven 3.9.1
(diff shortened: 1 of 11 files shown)


horstmannmat:
Ja, I think it makes sense to me, I was trying to follow the previous patterns, but it looks good :)
TYSM

PR #771, opened 2025-01-29
fix(API): pass translation_key_ids when removing keys from job

theSoenke:
Fix passing the `translation_key_ids` as arrays in query params are not properly supported by the generated client and move the param to the body. Also move the branch param to the request body to align it with keys_create`

https://github.com/phrase/phrase-cli/issues/158

### DIFF
--- a/paths/jobs/remove_keys.yaml
+++ b/paths/jobs/remove_keys.yaml
@@ -41,7 +41,7 @@ requestBody:
         type: object
         title: job/keys/delete/parameters
         required:
-          - name
+          - translation_key_ids
         properties:
           branch:
             description: specify the branch to use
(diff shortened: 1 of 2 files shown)


paths/jobs/remove_keys.yaml
```diff
@@ -46,6 +31,26 @@ x-code-samples:
...
-x-cli-version: '2.5'
+requestBody:
+  required: true
+  content:
+    application/json:
+      schema:
+        type: object
+        title: job/keys/delete/parameters
+        required:
+          - name
```
jablan, line 44: what is `name`?

jablan:
I'd also call it `fix(API)` rather than `fix(CLI,go)`

paths/jobs/remove_keys.yaml
theSoenke: thanks for checking. should be `translation_key_ids`

jablan (approved):

PR #772, opened 2025-01-29
feat(API): Adjust job_template_locale properties

BlumMichael:
Global job templates are not tied to a specific locale, so the locale `id` in the `job_template_locale` response can be `null`. Rather than modifying the existing `locale_preview` schema, this adds a new `locale_preview_global` schema with a nullable `id` field and uses it for the `locale` property in `job_template_locale`.

See:
- https://github.com/phrase/phrase/pull/14638
- https://phrase.atlassian.net/browse/STRINGS-1270

### DIFF
--- a/doc/compiled.json
+++ b/doc/compiled.json
@@ -3505,7 +3505,9 @@
         "title": "job_template_locale",
         "properties": {
           "id": {
-            "type": "string"
+            "type": "string",
+            "nullable": true,
+            "description": "The ID of the job template locale. Will return `null` if the related `job_template` is global."
           },
           "job_template": {
             "type": "object",
--- a/doc/compiled.json
+++ b/doc/compiled.json
@@ -1472,6 +1472,28 @@
           "code": "en-GB"
         }
       },
+      "locale_preview_global": {
+        "type": "object",
+        "title": "locale_preview_global",
+        "properties": {
+          "id": {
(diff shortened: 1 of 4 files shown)


doc/compiled.json
```diff
@@ -3279,7 +3279,9 @@
         "title": "job_template_locale",
         "properties": {
           "id": {
-            "type": "string"
+            "type": "string",
+            "nullable": true,
+            "description": "The ID of the job template locale. Will return `null` if the related `job_template` is global."
```
malvarezphrase, line 3284: **Issue**: If I'm not mistaken, we need to change the `locale_preview` schema so it can also just show the fields `name` and `code`. That would be in line `1433`. But maybe it's better to create a new schema because the field `id` won't be showing at all.

jablan: I agree, I'd vote for a separate schema

schemas/job_template_locale.yaml
```diff
@@ -5,10 +5,12 @@ job_template_locale:
   properties:
     id:
       type: string
+      nullable: true
```
jablan, line 8: hm is this also nullable, this is not a locale id, but the id of the "relationship" resource?

jablan (approved):
This is altering the API so it has to be `feat` and all the clients need a version bump.

PR #777, opened 2025-02-03
feat(API): Add source last updated at information on job details

forelabs:
Add source_translations_updated_at timestamp to jobs, showing last updated at from translations of job keys in the given source language.

Any affected ticket/issue:

- https://phrase.atlassian.net/browse/STRINGS-719

### DIFF
--- a/doc/compiled.json
+++ b/doc/compiled.json
@@ -3057,7 +3057,7 @@
               "job_tag_name": {
                 "type": "string"
               },
-              "source_updated_at": {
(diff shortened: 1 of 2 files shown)


jablan:
@forelabs this is not `docs(API)` change, but a `feat(API)`: it should bump minor versions of every library as there is an API change.

malvarezphrase (approved):
I think after adding Mladen's change in the title we should be fine ^^

malvarezphrase (approved):

PR #799, opened 2025-02-21
feat(API): add updated_since filter to job list #STRINGS-1555

jablan:
For https://phrase.atlassian.net/browse/STRINGS-1555 we need to reduce the volume of the list of jobs queried. This enables us to narrow down the list only to recently updated jobs.

### DIFF
--- a/parameters.yaml
+++ b/parameters.yaml
@@ -260,4 +260,31 @@ q:
   in: query
   schema:
     type: string
-
+job_owner:
+  description: filter by user owning job
+  example: abcd1234cdef1234abcd1234cdef1234
+  name: owned_by
+  in: query
+  schema:
+    type: string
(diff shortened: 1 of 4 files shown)


parameters.yaml
```diff
@@ -260,4 +260,31 @@ q:
...
+    type: string
+job_assignee:
+  description: filter by user assigned to job
+  example: abcd1234cdef1234abcd1234cdef1234
+  name: assigned_to
+  in: query
+  schema:
+    type: string
+job_state:
+  description: filter by state of job Valid states are <code>draft</code>, <code>in_progress</code>, <code>completed</code>
```
theSoenke, line 278: Believe there should be a dot after "job"

theSoenke (approved):

PR #809, opened 2025-02-25
fix(CLI): Revert viper library version #STRINGS-1487

jablan:
viper library (which we use for parsing yaml configuration) started [downcasing everything](https://github.com/spf13/viper?tab=readme-ov-file#does-viper-support-case-sensitive-keys) at one point, so the locale names are downcased whenever passed to Phrase API. that prevents them to be matched with their original names, causing slightly incorrect error about whole parameter `locale_mapping` missing.

until we see how to solve this properly (perhaps stop using viper at one point?), this PR rolls back to the working viper version.


theSoenke (approved):
interesting this is not supported by viper

PR #810, opened 2025-03-03
feat(CLI): Support cleanup over multiple sources #STRINGS-1378

jablan:
As our backend now supports multiple upload ids when deleting unmentioned keys, this PR uses that feature to simplify cleanup and enable cleanup over multiple sources.

Addresses https://github.com/phrase/phrase-cli/issues/147

### DIFF
--- a/clients/cli/cmd/internal/push.go
+++ b/clients/cli/cmd/internal/push.go
@@ -30,6 +30,12 @@ type PushCommand struct {
 	Tag                string
 }
 
+type PushResult struct {
+	ProjectID string
+	Branch    string
+	UploadIDs []string
+}
+
 func (cmd *PushCommand) Run() error {
 	if cmd.Config.Debug {
 		// suppresses content output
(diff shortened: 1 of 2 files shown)


david-redano:
Hi @jablan !

awesome thanks for the fix! I have a question though: is this considering key prefixes as well? Something like:

```
push:
    sources:
      - file: "./src/i18n/translations/reportengine-service/en-GB.json"
        params:
          tags: reportengine-service
          translation_key_prefix: reportengine-service.
          # filter_by_prefix: true
          locale_id: "en-GB"
          file_format: i18next
          # Add format options for i18next
          format_options:
            namespace: "reportengine-service"
          update_translations: true
pull:
 - file: "./src/i18n/translations/transaction-service/es-ES.json"
        params:
          tags: transaction-service
          translation_key_prefix: transaction-service.
          filter_by_prefix: true
          locale_id: "es-ES"
          file_format: i18next
          format_options:
            namespace: "transaction-service"
```
the same key would be prefixed with the prefix, so actually this would technically be different keys so that ideally should not be removed from the pull section files?

jablan:
hey @david-redano I can't say with 100% certainty without deeper looking into how prefixes work, but I suspect that they wouldn't affect the "combined cleanup" - if a key was mentioned in any of the sources, it should be preserved. technically these should be different sets of keys, but as long as they are mentioned in any of the sources, they should be preserved. anyway, soon you'll be able to try it out! :wink:

david-redano:
> hey @david-redano I can't say with 100% certainty without deeper looking into how prefixes work, but I suspect that they wouldn't affect the "combined cleanup" - if a key was mentioned in any of the sources, it should be preserved. technically these should be different sets of keys, but as long as they are mentioned in any of the sources, they should be preserved. anyway, soon you'll be able to try it out! 😉

Thanks for the reply @jablan ! Is there any chance that you or anyone else takes a `deeper looking into how prefixes work`? 
At Sage we are very interested in this tow work properly for our Use Cases.

Thanks again!

jablan:
hey @david-redano the prefix behaviour is explained here: https://support.phrase.com/hc/en-us/articles/5808300599068-Using-the-CLI-Strings#translation-key-prefix-0-10
so, upon upload, the prefix would be appended to the key name and that's (so, with prefixes) how you will see your keys in Phrase Strings UI. the particular upload is then "connected" with all the keys it contained and you can use `cleanup` option to delete all the keys that weren't part of an upload. this PR makes this working over multiple uploads (i.e. over multiple push sources) so that only the keys not mentioned in any of the uploads would be deleted.
if multiple uploaded files contain different prefixes, these would result with different keys, and all of them would be preserved.

PR #815, opened 2025-03-04
build(CLI): Build CLI app in CI [STRINGS-536] 

bikmazefe:
Build CLI app in CI to avoid introducing changes that crash the Go build.

https://phrase.atlassian.net/browse/STRINGS-536

### DIFF
--- a/.github/workflows/test-cli.yml
+++ b/.github/workflows/test-cli.yml
@@ -26,9 +26,12 @@ jobs:
           GOPRIVATE: github.com/phrase/phrase-go
         run: |
           npm install
+          pwd && ls -la
           cd ./clients/cli
+          pwd && ls -la
           go mod edit -replace github.com/phrase/phrase-go/v4=../go
           cd ../..
+          pwd && ls -la
           npm run generate.cli
           go build ./clients/cli/
       - name: License check
--- a/.github/workflows/test-cli.yml
+++ b/.github/workflows/test-cli.yml
@@ -27,12 +27,10 @@ jobs:
         run: |
           npm install
           npm run generate.go
-          ls -la clients/go
           cd ./clients/cli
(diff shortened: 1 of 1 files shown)


.github/workflows/test-cli.yml
```diff
@@ -16,12 +17,23 @@ jobs:
...
+          npm install
+          npm run generate.go
       - name: Generate CLI
         env:
           GOPRIVATE: github.com/phrase/phrase-go
         run: |
           npm install
+          cp ./clients/cli/go.mod ./clients/cli/go.mod.bak
+          cp ./clients/cli/go.sum ./clients/cli/go.sum.bak
+          echo "replace github.com/phrase/phrase-go => ../go" >> ./clients/cli/go.mod
```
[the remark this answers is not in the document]
bikmazefe: Ah yeah, I was thinking that we'd need to reset the `go.mod` into its original state but that won't be necessary I guess 👍 Thanks!

.github/workflows/test-cli.yml
```diff
@@ -21,7 +22,11 @@ jobs:
           GOPRIVATE: github.com/phrase/phrase-go
         run: |
           npm install
+          npm run generate.go
+          cd ./clients/cli
+          go mod edit -replace github.com/phrase/phrase-go/v4=../go
```
jablan, line 27: I don't understand how this works before the cli is actually generated (I would put `npm run generate.go` and `generate.cli` next to each other), but if it works as is, fine. :smile:

jablan (approved):

.github/workflows/test-cli.yml
bikmazefe: I think `go.mod` is static for CLI  🤔

PR #822, opened 2025-04-07
feat(CLI): Do not save token to config file by default #STRINGS-999

jablan:
When running `phrase init`, we should not save the token to the configuration file by default, as that's not safe from security perspective (tokens should not be committed to the repository).

Instead, we ask the user explicitly if they want to save it, otherwise we don't.

Addresses https://github.com/phrase/phrase-cli/issues/18

### DIFF
--- a/clients/cli/cmd/internal/init.go
+++ b/clients/cli/cmd/internal/init.go
@@ -143,9 +143,9 @@ func (cmd *InitCommand) askForToken() error {
 }
 
 func (cmd *InitCommand) setToken(token string) {
-	pushNow := ""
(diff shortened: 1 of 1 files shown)


clients/cli/cmd/internal/init.go
```diff
@@ -143,7 +143,11 @@ func (cmd *InitCommand) askForToken() error {
 }
 
 func (cmd *InitCommand) setToken(token string) {
-	cmd.YAML.AccessToken = token
+	saveTokenToConfig := ""
+	_ = prompt.WithDefault("Do you want to save the token to the config file? Otherwise you will have to provide it every time you run the command either through a command argument or through an environment variable (y/n)", &saveTokenToConfig, "n")
```
theSoenke, line 147: @jablan could you mention the argument and environment variable? might make it a little easier when it already shows you how to pass it

PR #834, opened 2025-05-22
feat(API): Add omit_statistics param to tags#show

forelabs:
- Document `omit_statistics` param on tags#show

### DIFF
--- a/doc/compiled.json
+++ b/doc/compiled.json
@@ -19411,6 +19411,15 @@
             "$ref": "#/components/parameters/name"
           },
           {
+            "description": "omit statistics in the response",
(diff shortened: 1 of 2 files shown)


paths/tags/index.yaml
```diff
@@ -14,6 +14,12 @@ parameters:
     in: query
     schema:
       type: boolean
+  - description: omit statistics in the response
```
jablan, line 17: should it be added also to show endpoint?

docstun (changes requested):
This parameter is also supported in the show action, but here it is only documented for the list (index) action.

docstun (approved):

PR #846, opened 2025-06-10
feat(API): Add use_ordinal_rules boolean [STRINGS-2273]

ildarsafin:
`use_ordinal_rules` boolean is used to define whether is using ORDINAL or CARDINAL pluralization rules.

Add `ordinal_plural_forms` to locales endpoint.

### DIFF
--- a/schemas/locale.yaml
+++ b/schemas/locale.yaml
@@ -44,6 +44,9 @@ locale:
     - zero
     - one
     - other
+    plural_ordinal_forms:
+      - zero
+      - other
     source_locale:
       id: abcd1234cdef1234abcd1234cdef1234
       name: en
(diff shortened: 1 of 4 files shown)


schemas/translation_key.yaml
```diff
@@ -13,6 +13,8 @@ translation_key:
       type: string
     plural:
       type: boolean
+    use_ordinal_rules:
```
jablan, line 16: perhaps also add in the example below?

forelabs: yes I think it should be added there as well

schemas/locale.yaml
```diff
@@ -40,6 +44,9 @@ locale:
     - zero
     - one
     - other
+    plural_ordinal_forms:
```
jablan, line 47: should be `ordinal_plural_forms` instead, right?

jablan (approved):

PR #848, opened 2025-06-16
docs(API): Convert descriptions to Markdown #STRINGS-2424

lookasc:
### Description

Converted HTML embedded in description fields into Markdown format, which is supported by Mintlify and the new Developer Hub.

Issue example:

Screenshot 2025-06-16 at 16 27 14

### Related issue
- [STRINGS-2424: Strings docs: Fix HTML in param description](https://phrase.atlassian.net/browse/STRINGS-2424)


jablan:
@lookasc the PR title should be `docs(API): ... #STRINGS-2424` as this does not affect the clients and shouldn't result with a version increase.

parameters.yaml
```diff
@@ -275,7 +275,7 @@ job_assignee:
   schema:
     type: string
 job_state:
-  description: "filter by state of job; valid states are: <code>draft</code>, <code>in_progress</code>, <code>completed</code>"
+  description: 'filter by state of job; valid states are: <code>draft</code>, <code>in_progress</code>, <code>completed</code>'
```
jablan, line 278: HTML still present here. also, would be nice to avoid unrelated lint changes (double to single quotes, indentation etc). not sure how to enforce a consistent style that should be respected by all our editors, but there should be a way.

lookasc: @jablan You're totally right about unrelated lint changes. These have been accidentally done by VSCode's `Save all` action. That's why this PR is still a `draft` 🙂

main.yaml
```diff
@@ -3,8 +3,7 @@ openapi: 3.0.3
 info:
   title: Phrase Strings API Reference
   version: 2.0.0
-  description:
-    $ref: "./intro.md"
+  description: Phrase Strings is a translation management platform for software projects. You can collaborate on language file translation with your team or order translations through our platform. The API allows you to import locale files, download locale files, tag keys or interact in other ways with the localization data stored in Phrase Strings for your account.
```
lookasc, line 6: I removed `intro.md` to avoid giving the impression that it will be handled by Mintlify.
All introduction-like pages should be edited in the [phrase/developer-hub](https://github.com/phrase/developer-hub/tree/main/docs/en/api/strings) repository.

lookasc: @jablan What exactly do you mean? All these pages are created from `intro.md`.
Screenshot 2025-06-17 at 17 05 41

jablan: but `intro.md` is deleted in this PR?

headers.yaml
```diff
@@ -19,7 +19,7 @@ X-Rate-Limit-Remaining:
   schema:
     type: integer
 X-Rate-Limit-Reset:
-  description: Timestamp of end of current time period as UNIX timestamp, see [Rate Limiting](#overview--rate-limiting)
+  description: Timestamp of end of current time period as UNIX timestamp, see [Rate Limiting](/en/api/strings/pagination#rate-limiting)
```
jablan, line 22: are we in control of the resulting markup? could we use `<base>` tag to specify `/en/api/strings/` as the base url and then be able to use just relative urls in all internal links?

main.yaml
jablan: ah sorry, I missed the fact that the content has been copied to https://github.com/phrase/developer-hub/tree/main/docs/en/api/strings

headers.yaml
lookasc: Unfortunately, it's not possible with Mintlify. This is a limitation that we have to live with at least until Mintlify provides a better way of handling languages.
The only drawback is that the user will always be redirected to `/en` regardless of the currently selected language.

clients/cli/go.mod
```diff
@@ -26,20 +26,16 @@ require (
 	github.com/spf13/cast v1.7.1 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
 	github.com/subosito/gotenv v1.6.0 // indirect
-	golang.org/x/exp v0.0.0-20241217172543-b2144cdd0a67
```
jablan, line 29: I'd suggest removing these two files from the PR. Pretty sure this was not caused by the rest of the changes.

headers.yaml
jablan: ok, that's also something that we can improve in the future.

clients/cli/go.mod
lookasc: Sure. Removed.

PR #869, opened 2025-07-14
docs(CLI): Add NOTICE.txt

forelabs:
Adding NOTICE.txt


forelabs:
Generation is documented in confluence. Keeping up to date will be a follow up that is not yet defined.

PR #873, opened 2025-07-23
chore(main): Remove docs

docstun:
No longer needed since the switch to the new [developer hub](https://developers.phrase.com/).

### Todo

- [ ] Update NOTICE
- [ ] Remove Netlify connection
- [x] Update UML (SVG)


docstun:
@jablan good point. Developer Hub supports staging/preview deployments, is that not sufficient?

docstun:
> @docstun that's not the same. new developer hub supports starting local server, one needs to set up the connection (IIRC I was doing it using symlinks), but that info needs to live in developer hub docs anyway, cause it's not open source. I'll open a PR there.

I was referring to the fact that any new branch in developer hub will spin up a preview deployment that would let you review the doc changes there under a dedicated URL which is very convenient for reviewing and sharing (but yes, it does require another repo besides this one).

jablan:
https://github.com/phrase/developer-hub/pull/26

I meant rather seeing local changes before they are committed and pushed at all (similar to normal local development process).

docstun:
> [phrase/developer-hub#26](https://github.com/phrase/developer-hub/pull/26)
> 
> I meant rather seeing local changes before they are committed and pushed at all (similar to normal local development process).

@jablan thanks for clarifying, I think you're right.

docstun:
@jablan can you re-create the workflow SVG? I do not know of have this tool.

jablan (approved):

PR #890, opened 2025-08-01
docs(API): Add examples in upload create params

Varpuspaavi:
Add examples in upload create params


jablan:
@hahmed-dev which kind of warning you have in mind? perhaps it could be mentioned in here https://developers.phrase.com/en/api/strings/uploads/upload-a-new-file#body-update-custom-metadata

PR #892, opened 2025-08-06
feat(API): Add support for  option on upload

hahmed-dev:
Add support for `update_custom_metadata` option on upload
Screenshot 2025-08-06 at 17 04 31

## Effected Ticket
https://phrase.atlassian.net/browse/SCM-619

## Related PR
https://github.com/phrase/phrase/pull/15791

### DIFF
--- a/clients/php/test/Api/UploadsApiTest.php
+++ b/clients/php/test/Api/UploadsApiTest.php
@@ -97,6 +97,7 @@ class UploadsApiTest extends TestCase
             null, # update_translations
             null, # update_custom_metadata
             null, # update_translation_keys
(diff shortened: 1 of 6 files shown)


clients/php/test/Api/UploadsApiTest.php
```diff
@@ -97,6 +97,7 @@ public function testUploadCreate()
             null, # update_translations
             null, # update_custom_metadata
             null, # update_translation_keys
+            null, # clear_custom_metadata
```
Varpuspaavi, line 100: I think this should be one line up so it aligns with the order

PR #912, opened 2025-09-18
feat: add branch sync endpoint

theSoenke:


doc/compiled.json
```diff
@@ -18301,16 +18301,89 @@
...
               }
             }
           }
+        }
+      }
+    },
+    "/projects/{project_id}/branches/{name}/sync": {
+      "patch": {
+        "summary": "Sync a branch",
+        "description": "Sync an existing branch.\n\n*Note: Only available for branches created with New branching.*\n",
```
forelabs, line 18318:
```suggestion
        "description": "Sync an existing branch.\n\n*Note: Only available for branches created with new branching.*\n",
```

also wondering if we should mention that it is in a private beta?

paths/branches/sync.yaml
```diff
@@ -0,0 +1,56 @@
+---
+summary: Sync a branch
+description: |
+  Sync an existing branch.
+
+  *Note: Only available for branches created with New branching.*
```
forelabs, line 6:
```suggestion
  *Note: Only available for branches created with new branching.*
```

forelabs (commented):
Was `New branching` defined like that or is it a typo?

paths/branches/sync.yaml
theSoenke: thanks, has been adjusted

jablan (approved):

PR #934, opened 2025-10-21
docs(API): mark target_locale_id required

theSoenke:
- Fix typo in exclude
- Mark target_locale_id as required


forelabs (approved):

jablan:
This is an API change, not only docs, should result with the version bump. Or?

theSoenke:
@jablan @forelabs i've removed required part again for now. In theory the endpoint could also be adjusted to work without the target_locale_id

PR #935, opened 2025-10-22
feat(API): Support source_locale_id param in uploads

bikmazefe:
https://phrase.atlassian.net/browse/SCD-408


forelabs (approved):
We need to release those new clients then :slightly_smiling_face:

PR #936, opened 2025-10-23
test(php): use named arguments in tests

jablan:
PHP supports named arguments from v8. We can use that in tests to avoid having to adapt them every time an argument is added to upload/download calls (which happens occasionally).


bikmazefe (approved):
🙌

PR #953, opened 2025-10-28
docs(API): Clarify the user field on translation schema

jablan:
https://memsource.slack.com/archives/C02QV9K8EP5/p1760016489742879


jablan:
@hahmed-dev I think it's not very realistic to add another field any time soon. I'd suggest to make this field behave consistently.

PR #966, opened 2025-11-13
feat(API): Add create comparison endpoint for branches [SCD-549]

bikmazefe:
Add create comparison endpoint for branches

https://phrase.atlassian.net/browse/SCD-549

### DIFF
--- a/doc/compiled.json
+++ b/doc/compiled.json
@@ -18462,7 +18462,7 @@
     "/projects/{project_id}/branches/{name}/sync": {
       "patch": {
         "summary": "Sync a branch",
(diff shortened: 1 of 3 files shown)


doc/compiled.json
```diff
@@ -18569,6 +18601,75 @@
...
+          },
+          {
+            "$ref": "#/components/parameters/project_id"
+          },
+          {
+            "$ref": "#/components/parameters/name"
+          }
+        ],
+        "responses": {
+          "200": {
```
jablan: does API ever return 200 here?

PR #978, opened 2025-12-03
feat(API): add pr_branch param to repo-sync export #SCD-594

BlumMichael:
add `pr_branch` param to repo-sync export

### DIFF
--- a/doc/compiled.json
+++ b/doc/compiled.json
@@ -4961,16 +4961,6 @@
         },
         "example": "my-feature-branch"
       },
-      "pr_branch": {
-        "name": "pr_branch",
-        "in": "query",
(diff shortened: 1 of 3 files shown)


doc/compiled.json
```diff
@@ -4886,6 +4886,16 @@
         },
         "example": "my-feature-branch"
       },
+      "pr_branch": {
+        "name": "pr_branch",
+        "in": "query",
+        "description": "Source branch to open a pull request from",
```
forelabs: Wondering if we should call it like that or use what we have in UI


```suggestion
        "description": "Branch name for the pull request. If left empty, a branch will be created with the default value",
```

@jablan wdyt?

PR #1085, opened 2026-04-03
fix(CLI): cache ignored when using --parallel with phrase pull

tetienne:
When splitting the work between #1066 (--cache) and #1067 (--parallel), I forgot to thread the cache through the parallel code path. As a result, `--cache --parallel` silently ignored the cache: no conditional headers were sent, no ETags were stored, and every run re-downloaded all locale files.

Tested against a production project with 204 locale files:

```
# First run — populates cache
$ ./phrase pull --cache --parallel
Downloaded en to locales/en/tradfile.json
...
(204 files downloaded)

# Second run — all 304s, nothing re-downloaded
$ ./phrase pull --cache --parallel
Not modified locales/en/tradfile.json
...
(204 files skipped)
```

## Test plan

- [x] `go build` compiles
- [x] Manual test: `phrase pull --cache --parallel` second run returns 304 for all files
- [x] `phrase pull --parallel` without `--cache` behaves identically to before
- [x] `phrase pull --cache` without `--parallel` unaffected


forelabs:
@tetienne We had de change our priorities a bit. We will now pick it up again. ~There is a conflict, can you take care of it please?~ Thank you a lot for your support.

clients/cli/cmd/internal/pull.go
```diff
@@ -285,54 +294,50 @@ func (target *Target) PullParallel(client *phrase.APIClient) error {
...
-	gate.RUnlock()
+func (target *Target) downloadWithRateGate(client *phrase.APIClient, localeFile *LocaleFile, opts phrase.LocaleDownloadOpts, gate *sync.RWMutex) (*os.File, *phrase.APIResponse, error) {
+	waitForGate := func() { gate.RLock(); gate.RUnlock() }
+	waitForGate()
 
 	file, response, err := client.LocalesApi.LocaleDownload(Auth, target.ProjectID, localeFile.ID, &opts)
 	if err != nil {
+		if response != nil && response.StatusCode == 304 {
+			return nil, response, nil
+		}
```
forelabs, line 322: A 304 is intercepted, the caller then re-checks` response.StatusCode == 304`. It works, but the responsibility boundary is blurry.

forelabs (approved):
ℹ️ Cache key changes for plain LocaleDownloadOpts fields (plain fields were previously omitted from the cache key)

Correctness fix, but it means any existing cache files with non-zero plain fields in LocaleDownloadOpts will produce different keys after upgrade. Not a big issue though, as this feature is new anyways.

PR #1099, opened 2026-04-13
feat(API): Add source_locale_id to job templates create/update

Varpuspaavi:
- Add source_locale_id to job templates create/update
- https://phrase.atlassian.net/browse/SCM-1327

### DIFF
--- a/paths/job_templates/create.yaml
+++ b/paths/job_templates/create.yaml
@@ -66,7 +66,7 @@ requestBody:
             type: boolean
             example: true
           source_locale_id:
-            description: The API id of the source language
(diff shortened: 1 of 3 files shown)


anthonyshchang (approved):

paths/job_templates/create.yaml
```diff
@@ -65,6 +65,10 @@ requestBody:
             description: Automatically translate the job using machine translation.
             type: boolean
             example: true
+          source_locale_id:
+            description: The API id of the source language
```
ildarsafin: **Suggestion** Could we explain here maybe that the passed ID of the project/account language will be set as a source language for the job template, otherwise the default language of the project will be used?

ildarsafin (approved):

PR #1125, opened 2026-05-06
chore(CLI): add cli config skill

theSoenke:
Add a Skill to help with creating a `.phrase.yml` config file

Still requires some documentation how to use and install it but that would be step 2


jablan:
where should this live and how is it accessed?

how do we maintain it, especially all the nitty gritty format option details? it's proven to be hard to maintain even in the HC.

forelabs:
Did you also consider how we can use this in our repo sync flow with the "1-button" solution?

PR #1127, opened 2026-05-13
feat(API): Add q search param to locales list endpoint

hahmed-dev:
## Summary

Adds optional q query parameter to the (List Locales) endpoint
Enables case-insensitive prefix filtering of tags by name (e.g. ?q=name:en)
Parameter type: string, optional — existing callers are unaffected

### DIFF
--- a/paths/locales/index.yaml
+++ b/paths/locales/index.yaml
@@ -21,6 +21,12 @@ parameters:
   in: query
   schema:
     type: string
+- description: Specify a query to filter locales by name.
+  example: name:en
+  name: q
+  in: query
+  schema:
+    type: string
 responses:
   '200':
     description: OK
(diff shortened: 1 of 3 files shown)


hahmed-dev:
Screenshot 2026-05-13 at 10 43 30

paths/locales/index.yaml
```diff
@@ -21,6 +21,12 @@ parameters:
   in: query
   schema:
     type: string
+- description: Specify a query to filter locales by name.
```
jablan:
```suggestion
- description: Specify a query to filter locales. Currently supports `name` argument, filtering only locales with names starting with the given string.
```

jablan (changes requested):
some more clarification of the `q` parameter

jablan (approved):

PR #1136, opened 2026-05-19
fix(API): branching: align operations with the backend behavior

forelabs:
Aligns the Branches operations and branch schema with the actual behavior of Api::V2::BranchesController:

- Add 401, 403, 422, and 409 responses where reachable across all Branches endpoints; 403 descriptions note the specific conditions (OAuth scope, Pundit, restrict_feature!(:branching))
- Document asynchronous semantics for create, merge, and sync: background jobs, transitional states, empty response bodies for merge/sync, and guidance to poll until state settles
- Add base field to create request body and branch schema (nullable, newer-branching system only)
- Declare strategy as an enum (use_main, use_branch) in merge and sync
- Enumerate the seven public-facing state values in branch schema
- Note that the Branching account feature must be enabled on every operation; clarify that update only accepts name

### DIFF
--- a/doc/compiled.json
+++ b/doc/compiled.json
@@ -3631,6 +3631,11 @@
           "name": {
             "type": "string"
           },
+          "base": {
+            "type": "string",
+            "nullable": true,
+            "description": "Name of the base branch this branch was created from. Only present for branches created with the newer branching system."
+          },
           "created_at": {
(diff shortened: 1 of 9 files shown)


jablan (approved):
I altered the title a bit.

PR #1148, opened 2026-06-01
feat(API): add fallback_for_unverified_translations param to locale d…

BlumMichael:
…ownload endpoints

Adds the new boolean parameter `fallback_for_unverified_translations` (default `false`) to:
- GET /api/v2/projects/{project_id}/locales/{id}/download
- POST /api/v2/projects/{project_id}/locales/{locale_id}/downloads

Also updates the `fallback_locale_id` and `use_locale_fallback` descriptions on both operations to reflect the relaxed constraint: `include_empty_translations=true` is no longer required when `fallback_for_unverified_translations=true` is set.

Corresponds to strings-app PR #17730 (SCD-950).

### DIFF
--- a/clients/java/src/test/java/com/phrase/client/api/LocalesApiTest.java
+++ b/clients/java/src/test/java/com/phrase/client/api/LocalesApiTest.java
@@ -168,6 +168,7 @@ public class LocalesApiTest {
         Boolean useLastReviewedVersion = null;
         String fallbackLocaleId = null;
         Boolean useLocaleFallback = null;
+        Boolean fallbackForUnverifiedTranslations = null;
         String sourceLocaleId = null;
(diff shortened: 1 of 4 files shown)


jablan:
oh we have an agent for this https://github.com/Phrase-Engineering/strings-app/blob/main/.github/actions/openapi-sync/action.yml

BlumMichael:
> oh we have an agent for this https://github.com/Phrase-Engineering/strings-app/blob/main/.github/actions/openapi-sync/action.yml

@jablan Can this agent create a PR for `phrase/strings-openapi` ?

jablan:
@BlumMichael not yet, and that step would have to be manual so that we don't unintentionally expose something we shouldn't

jablan (approved):

PR #1151, opened 2026-06-04
feat(API): document branch compare output

theSoenke:
Document compare endpoint output


jablan (approved):
@theSoenke shouldn't this be `feat(API)`? it's changing the output of an endpoint

theSoenke:
> Sönke Behrendt (@theSoenke) shouldn't this be `feat(API)`? it's changing the output of an endpoint

True, missed that when writing the title

jablan (approved):

PR #1182, opened 2026-06-15
docs(API): improve post /projects/{project_id}/screenshots documentation

sbl:
Improves the documentation for `post /projects/{project_id}/screenshots`: sharper descriptions, parameter docs, error responses, and usage examples.

Closes DEVEX-115

### DIFF
--- a/paths/screenshots/create.yaml
+++ b/paths/screenshots/create.yaml
@@ -1,6 +1,13 @@
 ---
 summary: Create a screenshot
-description: Create a new screenshot.
+description: |
+  Creates a screenshot in a project to provide visual context for in-context translation. Attach translation keys to regions of the uploaded image so translators can see where each string appears in your UI.
+
+  This endpoint accepts a multipart/form-data request with a binary file upload, unlike most Phrase API endpoints that use JSON. Use a multipart form client or the -F flag in curl rather than a JSON body.
+
+  The screenshot name must be unique within the project (case-insensitive). When name is omitted, it is derived from the uploaded filename. The account must have the Screenshots feature enabled; requests to projects on accounts without it return 403.
+
+  Requires the write scope and manage permission on the project.
 operationId: screenshot/create
 tags:
 - Screenshots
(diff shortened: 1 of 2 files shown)


paths/screenshots/create.yaml
```diff
@@ -1,6 +1,13 @@
...
 summary: Create a screenshot
-description: Create a new screenshot.
+description: |
+  Creates a screenshot in a project to provide visual context for in-context translation. Attach translation keys to regions of the uploaded image so translators can see where each string appears in your UI.
+
+  This endpoint accepts a multipart/form-data request with a binary file upload, unlike most Phrase API endpoints that use JSON. Use a multipart form client or the -F flag in curl rather than a JSON body.
+
+  The screenshot name must be unique within the project (case-insensitive). When name is omitted, it is derived from the uploaded filename. The account must have the Screenshots feature enabled; requests to projects on accounts without it return 403.
+
+  Requires the write scope and manage permission on the project.
```
theSoenke: It's a bit unclear that it references the API token. Not sure whether it does make sense to add here or then we'd need to add it for every endpoint to document the scopes

paths/screenshots/create.yaml
```diff
@@ -23,56 +38,77 @@ responses:
...
     multipart/form-data:
       schema:
         type: object
         title: screenshot/create/parameters
+        required:
+          - filename
         properties:
           branch:
-            description: specify the branch to use
+            description: Branch to use instead of the project default.
```
theSoenke: It's somewhat unclear from the description what project default would mean. Think we should use a consistent description of the branch param across all endpoints

paths/screenshots/create.yaml
sbl: Agreed — rather than document scopes per endpoint, I dropped the 'Requires the write scope and manage permission' sentence from the description. The 403 response keeps the scope/permission explanation where it's actionable. No other endpoint documents scopes in prose, so this keeps it consistent.

paths/screenshots/create.yaml
sbl: Fixed: the branch property now reads 'specify the branch to use', matching the wording used across the other endpoints, and dropping the unclear 'instead of the project default'.

jablan (approved):

PR #1178, opened 2026-06-10
docs(API): improve GET /projects/:id/documents documentation

sbl:
Improves the documentation for `GET /projects/:project_id/documents` (list documents).

## Changes
- Adds conceptual description: what the operation does, why/when to use it, pagination model via Link header (RFC 5988), and pitfalls
- Documents 401 response
- Improves 403 response description
- Fixes curl example: was using `?project_id=asdf` query param; now correctly uses `:project_id` path param format
- Adds response example with realistic document values
- Improves `q` parameter description

All changes are spec-layer only (paths/documents/index.yaml + compiled bundle). No Ruby source edits.

Part of [DEVEX-119](https://phrase.atlassian.net/browse/DEVEX-119).

Drafted with AI assistance and grounded in the strings-app source. Please review for technical accuracy before merging; nothing is merged automatically.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[DEVEX-119]: https://phrase.atlassian.net/browse/DEVEX-119?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1183, opened 2026-06-15
docs(API): improve get /projects/{project_id}/documents documentation

sbl:
Improves the documentation for `get /projects/{project_id}/documents`: sharper descriptions, parameter docs, error responses, and usage examples.


sbl:
This PR replaces #1178, which was closed after reviews from jablan and theSoenke.

**Feedback addressed from #1178:**

| Reviewer | Comment | Status |
|---|---|---|
| theSoenke | "The policy should not be part of the API docs" | ✅ Fixed |
| theSoenke | "`per_page` and rate limit details do not need to be repeated per endpoint" | ✅ Fixed |
| jablan | "I don't think it's helpful to say the list is paginated here" | ✅ Fixed |
| jablan | "Pagination information is superfluous here" (RFC 5988 / Link header sentence) | ✅ Fixed |
| jablan | "Please do not expose identifier names from internal applications" | ✅ Fixed |
| jablan | "Rate limiting applies universally and is explained on a dedicated place" | ✅ Fixed |

**What changed:** The description is fully rewritten. All six flagged items are removed:
- No `DocumentPolicy#list_documents?` or any internal class/policy name
- No pagination prose ("returns a paginated list", RFC 5988, `Link` header)
- No `per_page` cap or default
- No rate-limit pitfall

What remains is endpoint-specific: what a document is, when you'd use this endpoint, how the `q` prefix match works, and the auth requirement.

paths/documents/index.yaml
```diff
@@ -1,19 +1,26 @@
 ---
 summary: List documents
-description: List all documents the current user has access to.
+description: |
+  Returns all documents in a project that the authenticated user has read access to. A Document is a source file — such as an HTML or YAML file — that has been uploaded to Phrase Strings and whose content is segmented into translation keys for localization.
```
theSoenke: Only HTML and docx files are document files that will be listed here

paths/documents/index.yaml
```diff
@@ -1,19 +1,26 @@
...
+
+  Use this endpoint to enumerate documents before downloading, previewing, or triggering translation workflows for individual files.
+
+  The q parameter performs a prefix match on the document name (case-insensitive). For example, passing q=invoice returns documents whose names begin with "invoice" but not documents containing "invoice" elsewhere in the name.
+
+  Requires the read OAuth scope. Returns 403 when the requesting user does not have read permission on documents in this project.
```
theSoenke: `Requires the read OAuth scope`. We should list the access token scopes consistently or drop them here

paths/documents/index.yaml
```diff
@@ -44,14 +63,30 @@ responses:
...
+        "created_at": "2015-01-28T09:52:53Z",
+        "updated_at": "2015-03-04T11:22:30Z"
+      },
+      {
+        "id": "ef012345ef012345ef012345ef012345",
+        "name": "invoice.yaml",
+        "created_at": "2015-02-10T14:30:00Z",
+        "updated_at": "2015-03-01T08:00:00Z"
+      }
+    ]
```
theSoenke: Responses should not be listed as code examples and we have examples above already

paths/documents/index.yaml
sbl: Fixed: the description now says a document is 'an HTML or DOCX file' (dropped YAML), and the example value invoice.yaml is now invoice.docx to match.

paths/documents/index.yaml
sbl: Dropped: removed the 'Requires the read OAuth scope' sentence from the description. No other endpoint documents scopes in prose, so going with 'drop them here'. The 403 response still explains the read requirement where it's actionable.

paths/documents/index.yaml
sbl: Removed the 'Response' code sample. The 200 example above already shows the response shape.

jablan (approved):

PR #1177, opened 2026-06-10
docs(API): improve DELETE /projects/:id/documents/:id documentation

sbl:
Improves the documentation for `DELETE /projects/:project_id/documents/:id` (delete document).

## Changes
- Adds conceptual description: what the operation does, why/when to use it (cleanup), the permanent-deletion model, and pitfalls
- Documents 401 and improves 403 response with causes and remediation (scope + manage permission requirement)
- Adds 422 response for model-level constraint failures
- Improves curl example to show expected HTTP 200 response

All changes are spec-layer only (paths/documents/destroy.yaml + compiled bundle). No Ruby source edits.

Part of [DEVEX-119](https://phrase.atlassian.net/browse/DEVEX-119).

Drafted with AI assistance and grounded in the strings-app source. Please review for technical accuracy before merging; nothing is merged automatically.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[DEVEX-119]: https://phrase.atlassian.net/browse/DEVEX-119?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1185, opened 2026-06-15
docs(API): improve delete /projects/{project_id}/documents/{id} documentation

sbl:
Improves the documentation for `delete /projects/{project_id}/documents/{id}`: sharper descriptions, parameter docs, error responses, and usage examples.


sbl:
This PR replaces #1177, which was closed after jablan's review.

**Feedback addressed from #1177:**

| Comment | Status |
|---|---|
| "`numeric primary key` is an internal development detail and should not be exposed" | ✅ Fixed |
| "This 403 description is almost fully generic and could be moved to the referenced section" | ✅ Fixed |

**What changed:**
- The description no longer mentions `numeric primary key` or `DocumentPolicy#destroy?`. Implementation details are translated into caller-facing behaviour: *"Requires the write scope on the access token and manage permission on the document."*
- The 403 description is now endpoint-specific: *"Obtain a token with the write scope or request manage access from a project administrator."* Generic 403 prose has been removed rather than moved inline.

paths/documents/destroy.yaml
```diff
@@ -10,31 +13,36 @@ parameters:
...
+      -X DELETE
 - lang: CLI v2
   source: |-
     phrase documents delete \
-    --project_id <project_id> \
-    --id <id> \
-    --access_token <token>
+    --project_id my-project-xyz123 \
+    --id doc-abc456 \
+    --access_token USERNAME_OR_ACCESS_TOKEN
```
theSoenke: Inconsistent that this uses the `access_token` param with other commands. And username wouldn't work

paths/documents/destroy.yaml
```diff
@@ -10,31 +13,36 @@ parameters:
...
-      -H 'Content-Type: application/json'
+      -X DELETE
 - lang: CLI v2
   source: |-
     phrase documents delete \
-    --project_id <project_id> \
-    --id <id> \
-    --access_token <token>
+    --project_id my-project-xyz123 \
+    --id doc-abc456 \
```
theSoenke: Ids look like slugs instead of API ids used in other endpoints

paths/documents/destroy.yaml
sbl: Fixed: the CLI v2 sample now uses `--access_token <token>` (a placeholder for an actual token, not a username), consistent with the other commands. The curl sample keeps `-u USERNAME_OR_ACCESS_TOKEN` since basic-auth accepts either there.

paths/documents/destroy.yaml
sbl: Fixed: replaced the slug-style values (`my-project-xyz123`, `doc-abc456`) with the `<project_id>` / `<id>` placeholders the other endpoints use.

jablan (approved):

PR #1186, opened 2026-06-15
docs(API): improve delete /projects/{project_id}/screenshots/{id} documentation

sbl:
Improves the documentation for `delete /projects/{project_id}/screenshots/{id}`: sharper descriptions, parameter docs, error responses, and usage examples.


paths/screenshots/destroy.yaml
```diff
@@ -1,6 +1,18 @@
...
+
+  Errors:
+
+  - screenshot_invalid_request (400): The request was malformed or a path parameter failed validation. Correct the parameter format and retry.
+  - screenshot_unauthorized (401): The access token is missing or invalid. Supply a valid access token.
+  - screenshot_forbidden (403): The access token lacks the write scope, the account plan does not include the screenshot feature, or the requesting user does not have manage permission on the screenshot. Upgrade your plan, add the write scope to your token, or request manage access.
+  - screenshot_not_found (404): No screenshot with this identifier exists in the project. Verify the project_id and id values.
+  - screenshot_unprocessable (422): The screenshot could not be deleted because the model reported validation errors. Inspect the errors array in the response body.
+  - screenshot_rate_limited (429): Too many requests have been made. Wait until the X-Rate-Limit-Reset time before retrying.
```
theSoenke: Should be listed as status code description and not part of the main description

paths/screenshots/destroy.yaml
```diff
@@ -17,32 +29,41 @@ parameters:
...
-    --branch my-feature-branch \
-    --access_token <token>
+      --project_id <project_id> \
+      --id <id> \
+      --access_token <token>
 x-cli-version: '2.5'
+security:
+  - Token: []
+  - OAuth2:
+      - write
```
theSoenke: This is new for this endpoint and not used yet in other endpoints. Either it should be used or consistently or not at all

paths/screenshots/destroy.yaml
```diff
@@ -17,32 +29,41 @@ parameters:
...
 - lang: Curl
   source: |-
     curl "https://api.phrase.com/v2/projects/:project_id/screenshots/:id" \
       -u USERNAME_OR_ACCESS_TOKEN \
-      -X DELETE \
-      -d '{"branch":"my-feature-branch"}' \
-      -H 'Content-Type: application/json'
+      -X DELETE
+
+    HTTP/1.1 204 No Content
```
theSoenke: Response should not be listed in the example cURL request

paths/screenshots/destroy.yaml
sbl: Done: removed the per-status `Errors:` enumeration (and the 403/404 sentences) from the top-level description. Each status code keeps its own description on the response object, so the per-code prose now lives only where you suggested.

paths/screenshots/destroy.yaml
sbl: Removed the endpoint-level `security` block entirely. No other endpoint declares one and the global `security: [Token, Basic]` in main.yaml already applies — the block was also silently overriding Basic with OAuth2 on just this operation, which is what oasdiff flagged. Going with 'not at all' for consistency.

paths/screenshots/destroy.yaml
sbl: Fixed: dropped the `HTTP/1.1 204 No Content` line so the curl sample shows only the request, matching the rest of the reference.

jablan (approved):

PR #1174, opened 2026-06-10
docs(API): improve POST /icu/skeleton documentation

sbl:
Improves the documentation for `POST /icu/skeleton` (ICU skeleton builder).

## Changes
- Adds conceptual description: what the endpoint does, why/when to use it, the ICU/CLDR model, and pitfalls
- Expands ICU (International Components for Unicode) and CLDR (Common Locale Data Repository) abbreviations at first use
- Documents 401 and 422 responses with causes and remediation text
- Improves 403 description with remediation guidance
- Adds `enum` constraint for `cldr_version` parameter (`legacy` | `cldr_41`)
- Improves all parameter descriptions for clarity
- Adds response example to the 200 response

All changes are spec-layer only (paths/icu/skeleton.yaml + compiled bundle). No Ruby source edits.

Part of [DEVEX-117](https://phrase.atlassian.net/browse/DEVEX-117).

Drafted with AI assistance and grounded in the strings-app source. Please review for technical accuracy before merging; nothing is merged automatically.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[DEVEX-117]: https://phrase.atlassian.net/browse/DEVEX-117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1187, opened 2026-06-15
fix(API): improve post /icu/skeleton documentation

sbl:
Improves the documentation for `post /icu/skeleton`: sharper descriptions, parameter docs, error responses, and usage examples.

Closes DEVEX-117.


sbl:
This PR replaces #1174, which was closed after jablan's review.

**Feedback addressed from #1174:**

| Comment | Status |
|---|---|
| "Moving existing sections around makes reviewing harder" | ✅ Fixed |

**What changed:** `requestBody` was placed at the bottom of the file in #1174 (after `x-code-samples`). This PR puts it back between `parameters` and `responses` — the standard OpenAPI order — and leaves all other section ordering untouched. Only the fields required by the failing rubric dimensions are modified.

paths/icu/skeleton.yaml
```diff
@@ -23,69 +141,78 @@ responses:
...
-      -d '{"content":"{number, plural, one {One} other {%{n}}}","locale_codes":["en"],"zero_form_enabled": true}' \
-      -H 'Content-Type: application/json'
+      -H "Content-Type: application/json" \
+      -d '{"content":"{count, plural, one {One item} other {# items}}","locale_codes":["en","de"],"zero_form_enabled":false}'
+
+    Example response:
+    {
+      "en": "{count, plural, offset:0 one {One item} other {# items}}",
+      "de": "{count, plural, offset:0 one {} other {}}"
+    }
```
theSoenke: The example is already covered above and this should only include the cURL request

paths/icu/skeleton.yaml
```diff
@@ -1,18 +1,136 @@
...
 parameters:
 - "$ref": "../../parameters.yaml#/X-PhraseApp-OTP"
+requestBody:
+  required: true
+  content:
+    application/json:
+      schema:
+        type: object
+        title: icu/skeleton/parameters
+        oneOf:
```
theSoenke, line 76: Both oneOf branches repeat all 6 property definitions — only required differs. Should be possible to move properties to the top level and combine with the oneOf constraint via allOf.

paths/icu/skeleton.yaml
sbl: Fixed in the latest push: dropped the `Example response` block from both the curl and CLI v2 samples, so each sample is now just the request — matching every other endpoint in the repo. The 200 example stays documented on the response object.

paths/icu/skeleton.yaml
sbl: Done: hoisted the six property definitions to the schema's top-level `properties` and reduced `oneOf` to just the `required` constraint (`[content]` xor `[id]`). The duplication is gone and the content-xor-id rule is unchanged.

sbl:
### API surface change — reviewer note

This PR constrains `cldr_version` on `POST /icu/skeleton` to `enum: [legacy, cldr_41]` — the only values the prior description already declared valid. oasdiff flags this as `request-property-became-enum`. No change to live API behavior; the enum encodes the documented contract.

Applying `breaking-change-approved` to clear the gate.

**Blocker (not cleared by the label):** this enum addition currently breaks the TypeScript client build (`jest` check):
`src/models/IcuSkeletonParameters.ts: error TS2304: Cannot find name 'IcuSkeletonParametersCldrVersionEnum'`.
The generator references the enum type but does not emit it. This must be resolved before merge (e.g. give the enum a named schema, or drop the `enum` and keep the allowed values in prose).

jablan (approved):

PR #1168, opened 2026-06-10
feat(API): improve get /projects/{project_id}/screenshots/{id} documentation

sbl:
Improves the documentation for \`get /projects/{project_id}/screenshots/{id}\`: sharper descriptions, parameter docs, error responses, and usage examples.

Drafted with AI assistance and grounded in the API implementation. Please review for technical accuracy before merging; nothing is merged automatically.

Jira: DEVEX-115

PR #1189, opened 2026-06-15
docs(API): improve get /projects/{project_id}/screenshots/{id} documentation

sbl:
Improves the documentation for `get /projects/{project_id}/screenshots/{id}`: sharper descriptions, parameter docs, error responses, and usage examples.


sbl:
This PR replaces #1168, which was closed after jablan's review.

**Feedback addressed from #1168:**

| Comment | Status |
|---|---|
| "Internal implementation detail and shouldn't be exposed" | ✅ Fixed |
| "No need to describe `id`/`branch` in the description — that belongs in `parameters`" | ✅ Fixed |
| "429 description is identical for all requests — move to `responses.yaml`" | ✅ Fixed |
| "It should be possible to hyperlink other endpoints" | ⚠️ Not addressed — see note below |

**What changed:**
- `branch` elaboration is moved to the parameter's own `description` field in `parameters:`. The top-level `description` no longer restates individual request fields.
- Internal detail removed from the description body.
- 429 now uses a bare `$ref: "../../responses.yaml#/429"` with no inline description override.

**Open item — hyperlinks:** Related endpoints are referenced in plain prose rather than as Markdown links. If clickable cross-references are a hard requirement, let us know and we'll update the description accordingly.

paths/screenshots/show.yaml
```diff
@@ -12,6 +15,7 @@ parameters:
   example: my-feature-branch
   name: branch
   in: query
+  required: false
```
jablan, line 18: branch parameter should be available in `parameters.yaml` to be used as a reference. we should double check if all current usages are non-required (most should be), and then add this flag there as well. perhaps as a separate PR

jablan (approved):

PR #1190, opened 2026-06-15
fix(API): improve patch /projects/{project_id}/screenshots/{id} documentation

sbl:
Improves the documentation for `patch /projects/{project_id}/screenshots/{id}`: sharper descriptions, parameter docs, error responses, and usage examples.


sbl:
### API surface change — reviewer note

This PR changes the request-body media type of `PATCH /projects/{project_id}/screenshots/{id}` from `application/json` to `multipart/form-data`, which is correct for an endpoint that accepts a binary `filename` upload (and matches `POST .../screenshots`). oasdiff flags `request-body-media-type-removed`. No change to live API behavior; it corrects the documented transport.

Applying `breaking-change-approved` to clear the gate.

doc/compiled.json
```diff
@@ -21616,6 +21616,41 @@
...
+                    "type": "string",
+                    "minLength": 1,
+                    "example": "home_screen_v2"
+                  },
+                  "description": {
+                    "description": "Updated description for the screenshot. Omit to leave the current description unchanged.",
+                    "type": "string",
+                    "example": "Updated landing page screenshot"
+                  },
+                  "filename": {
```
jablan, line 21649: I think the same parameter in `create` endpoint has different description, also some accepted formats (no example though).
we should try to keep the same descriptions for the same parameters across endpoints, or, even better, extract them to schemas and use as references

paths/screenshots/update.yaml
```diff
@@ -24,57 +55,14 @@ responses:
...
   '403':
     "$ref": "../../responses.yaml#/403"
-    description: Forbidden. Returned when the access token lacks the `write` scope, when the requesting user is not allowed to update this screenshot, or when the account does not have the Attachable Screenshots feature.
+  '404':
+    "$ref": "../../responses.yaml#/404"
   '422':
     "$ref": "../../responses.yaml#/422"
   '429':
     "$ref": "../../responses.yaml#/429"
-x-code-samples:
```
jablan, line 38: why are the samples removed?

sbl: they are automatically rendered by mintlify - this removes the duplicatoin and the risk of "example drift"

jablan: just took a look. wow, those automatic examples are wrong. they are not using our generated clients at all.

jablan: I mean, probably makes sense for `curl` and e.g. the languages we don't cover. but others should be using our clients. I also suspect it can't generate examples for the CLI?

sbl: > wow, those automatic examples are wrong
can you send an example and clarify what's wrong?  "in theory" examples are generated against our spec. So my suspicion would usually be if the spec is wrong the examples are wrong or something is missing.

I wouldn't augment the scope now to the cli.

We can make a call to just keep the existing examples untoched. Or try to figure out why they are rendering so wrong.

jablan: our openapi project generates the documentation for our API, generates the CLI to work with it, but also generates client libraries for several languages. so, for example in Python, to get the information on account, one would use:

```python
import phrase_api
from phrase_api.rest import ApiException

configuration = phrase_api.Configuration()
api_client = phrase_api.ApiClient(configuration)
api_instance = phrase_api.AccountsApi(api_client)
...
api_response = api_instance.account_show('ACCOUNT_ID')
```

(example adapted from our python docs: https://github.com/phrase/phrase-python#getting-started)

and get a parsed response, depending on the language, cast into an appropriate structure or a class instance.

while the mintlify-generated example just performs a HTTP request to our API:

```python
import requests
url = "https://api.phrase.com/v2/accounts/{id}"
headers = {"Authorization": "<api-key>"}
response = requests.get(url, headers=headers)
```

so, more or less, a plain curl-type request.

technically, the example is not wrong, but it kinda beats a big part of the strings-openapi purpose.

I suggest we get rid of the curl examples (as they can be properly generated by mintlify), but we'd need to keep CLI examples, and potentially even expand it to other platforms we cater for (basically exposing already auto-generated platforms documentation)

paths/screenshots/update.yaml
```diff
@@ -55,7 +24,7 @@ x-code-samples:
 requestBody:
   required: true
   content:
-    application/json:
+    multipart/form-data:
```
jablan, line 27: this is a pretty significant change. although I don't doubt that it's the correct value, I'm wary how it would reflect to the generated clients (especially CLI). I suggest starting with the `/create` endpoint instead, and adding a CLI spec along. we want to make sure the CLI doesn't break on this change.

PR #1192, opened 2026-06-15
docs(API): improve post /projects/{project_id}/keys/{id}/key_links documentation

sbl:
Improves the documentation for `post /projects/{project_id}/keys/{id}/key_links`: sharper descriptions, parameter docs, error responses, and usage examples.

Jira: DEVEX-116


paths/key_links/create.yaml
```diff
@@ -19,29 +22,82 @@ requestBody:
...
 
 responses:
   '201':
-    description: Created
+    description: Key link reference created.
     content:
       application/json:
         schema:
           "$ref": "../../schemas/key_link.yaml#/key_link"
+        example:
```
jablan: I think the example should be in the referenced schema file.

jablan (changes requested):

paths/key_links/create.yaml
sbl: Done: moved the example onto the shared schema (`schemas/key_link.yaml`) as a schema-level `example:` and dropped the inline one here, so create, list, and batch unlink all inherit one canonical example via the `$ref` instead of each restating it.

jablan (approved):

PR #1188, opened 2026-06-15
fix(API): improve get /projects/{project_id}/keys/{id}/key_links documentation

sbl:
Improves the documentation for `get /projects/{project_id}/keys/{id}/key_links`: sharper descriptions, parameter docs, error responses, and usage examples.

Drafted with AI assistance and grounded in the API implementation. Please review for technical accuracy before merging; nothing is merged automatically.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1193, opened 2026-06-15
fix(API): improve delete /projects/{project_id}/keys/{id}/key_links documentation

sbl:
Improves the documentation for `delete /projects/{project_id}/keys/{id}/key_links`: sharper descriptions, parameter docs, error responses, and usage examples.


paths/key_links/batch_destroy.yaml
```diff
@@ -18,28 +23,188 @@ requestBody:
...
+            - keep_content
+            - remove_content
+            default: keep_content
 responses:
   '200':
-    description: OK
+    description: Updated linked-key group reference after the unlink operation.
+    content:
+      application/json:
+        schema:
```
jablan, line 48: should be referenced from `/schemas`. I guess it's `/schemas/key_link.yaml` but needs to be double-checked

jablan (changes requested):

paths/key_links/batch_destroy.yaml
sbl: Done, and double-checked. The 200 now references `../../schemas/key_link.yaml#/key_link` — it's the same linked-key-group resource that create (#1192) and list already return, so the inlined `linked_keys_reference` object was a straight duplicate of that schema. The operation-level `example:` stays inline since it shows the post-unlink state (remaining children), which differs from the create example.

sbl: Correction to my note above: I've now also moved the example. Referencing the schema was only half the fix — the 200 still had an inline `example:` overriding the `$ref` (with off-convention slug-style ids). It now renders the canonical example from `schemas/key_link.yaml`, matching #1192 and #1188. No inline example remains here.

jablan: I think overriding the `$ref` doesn't even work.

jablan (approved):

PR #1194, opened 2026-06-15
fix(API): improve delete /projects/{project_id}/keys/{id}/key_links/{child_key_id} documentation

sbl:
Improves the documentation for `delete /projects/{project_id}/keys/{id}/key_links/{child_key_id}`: sharper descriptions, parameter docs, error responses, and usage examples.


jablan:
if we want to merge these PRs individually, we should probably go with `fix(...)` instead of `feat(...)` in cases like this, where only the error response is being updated. otherwise we'll have huge number of minor version bumps, one for every PR. keeping them at the fix level, we could also group them in a way.

sbl:
> if we want to merge these PRs individually, we should probably go with `fix(...)` instead of `feat(...)` in cases like this, where only the error response is being updated. otherwise we'll have huge number of minor version bumps, one for every PR. keeping them at the fix level, we could also group them in a way.

@jablan very useful insight! so the automation here is triggered by the PR title? I'd then manually fix here and update the prompt instructions.

forelabs:
@jablan would you consider this as "fix" ? I'd say this is just "docs" update - wdyt?

sbl:
On the `feat` vs `fix` vs `docs` question (thanks @jablan @forelabs): we've standardized the batch as follows —

- **`docs(API):`** for PRs that only change descriptions / examples and produce no compiled-API change (oasdiff reports `No changes to report`). These cause no version bump.
- **`fix(API):`** for PRs that do change the compiled API surface (added/removed properties, response media types, headers, etc.).

We're avoiding `feat` for this work so we don't get a minor bump per PR. This PR changes the 422 response (adds `errors/items/code` and adjusts the rate-limit headers), so it stays `fix`. The pure description/example PRs in the batch have been retitled to `docs`.

paths/key_links/destroy.yaml
```diff
@@ -10,24 +15,83 @@ parameters:
...
+        example:
+          message: Child key 1234abcd1234cdef1234abcd1234cdef is not linked to this parent key.
+          errors:
+            - resource: base
+              field: base
+              message: Child key 1234abcd1234cdef1234abcd1234cdef is not linked to this parent key.
+              code: KEY_IS_NOT_LINKED
   '429':
+    description: |
+      Too many requests. You have exceeded the rate limit. Wait until the time indicated in X-Rate-Limit-Reset before retrying.
```
jablan: As this is a generic universal response for 429, it should live in `responses.yaml#429`. We should try to keep the error section as generic as possible to avoid duplicating the same or very similar content.

paths/key_links/destroy.yaml
```diff
@@ -10,24 +15,83 @@ parameters:
...
+        "$ref": "../../headers.yaml#/X-Rate-Limit-Remaining"
+      X-Rate-Limit-Reset:
+        "$ref": "../../headers.yaml#/X-Rate-Limit-Reset"
+    content: {}
   '400':
+    description: |
+      Bad request. Returned when the request is malformed, such as unparseable JSON in the request body. Fix the request format and retry.
     "$ref": "../../responses.yaml#/400"
   '401':
+    description: |
```
jablan: same as for 429 error below, try to find a generic formulation and update `responses.yml#401` with it. if there are more than one generic case, we could separate it to `401_write` and `401_read` for example and use the appropriate reference where needed.

paths/key_links/destroy.yaml
sbl: Done. The inline 429 description is removed — this response is now a bare `$ref: ../../responses.yaml#/429`. The generic wording lives once in the shared response, enriched in #1208 (cause + remediation pointing at the `X-Rate-Limit-Reset` header). The same treatment was applied across the sibling error-doc PRs so the 429 text cannot drift between endpoints.

paths/key_links/destroy.yaml
sbl: Done. 401 is now a bare `$ref: ../../responses.yaml#/401`, with the generic formulation centralized in the shared response (#1208). On the write/read split: 401 is authentication only (token missing, expired, or invalid), which has no read-vs-write variant — that distinction is a 403 (scope/permission) concern, so I kept a single generic 401 and left scope nuance to 403. Happy to split into 401_write/401_read if you would prefer.

jablan (approved):

PR #1206, opened 2026-06-23
docs(API): gate multi-project automation on advanced_job_automation plan feature #SCM-1408

hahmed-dev:
## What changed

Update the `project_ids` field description in the automation **create** and **update** operations to reflect the current API contract:

- Multi-project support (more than one entry in `project_ids`) is now gated on the `advanced_job_automation` plan feature.
- Accounts without this feature receive a `422 Unprocessable Entity` response with error field `project_ids`.
- The previous description ("currently only the first ID in the array is used / leaves room for future support") was stale and no longer accurate.

## Affected operations

- `POST /api/v2/accounts/{account_id}/automations`
- `PATCH /api/v2/accounts/{account_id}/automations/{id}`

## Related

strings-app PR: https://github.com/Phrase-Engineering/strings-app/pull/17939 (SCM-1408)

### DIFF
--- a/paths/automations/create.yaml
+++ b/paths/automations/create.yaml
@@ -33,8 +33,9 @@ requestBody:
           project_ids:
             description: |
               List of project IDs to associate with the automation.
-              Currently, only the first ID in the array is used.
-              The array format leaves room for future support of multiple projects.
+              Providing more than one project ID requires the `advanced_job_automation`
+              plan feature; accounts without this feature receive a 422 response with
+              error field `project_ids`.
             type: array
             items:
               type: string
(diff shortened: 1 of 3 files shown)


paths/automations/create.yaml
```diff
@@ -33,8 +33,9 @@ requestBody:
           project_ids:
             description: |
               List of project IDs to associate with the automation.
-              Currently, only the first ID in the array is used.
-              The array format leaves room for future support of multiple projects.
+              Providing more than one project ID requires the `advanced_job_automation`
```
jablan, line 36: as this same parameter is duplicated in 2 endpoints, it could be extracted and referenced in schemas see https://github.com/phrase/strings-openapi/blob/ce2ed9488e111fb5d9bc3810a78c47d23553c8b7/paths/custom_metadata_properties/create.yaml#L26 for example

jablan (approved):

PR #1209, opened 2026-06-24
fix(cli): detect file upload params by type, and run CLI build on PRs

jablan:
## What

Two related fixes prompted by the `screenshot/create` endpoint breaking the CLI build (`go build` in `clients/cli` failed with `params.Get*os.File` and a `too many arguments` call).

### 1. Detect binary upload params by type, not by name

The CLI handlebars template detected file-upload params by matching the literal param name (`file`/`filename`). A **required** binary field with any other name fell through to the primitive handler and generated invalid Go:

```go
filename := params.Get * os.File(helpers.ToSnakeCase("Filename"))
```

`screenshot/create`'s binary field is named `filename` and is required, hitting exactly this case.

Both branches now use `{{#isFile}}` — the flag the generator derives from `format: binary`, and the same one the go/python/typescript/php templates in this repo already use. Future binary fields (`image`, `attachment`, …) work automatically.

### 2. Make CI actually compile the CLI on PRs

This break reached `main` because **no CI job compiled the CLI on the PR**:

- **`test-cli.yml`** is the only job that runs `go build` on the CLI, but its `paths:` filter excluded the spec (`paths/**`, `schemas/**`, shared `*.yaml`) and the CLI templates the CLI is generated from — so a spec-only change skipped it. Switched to `on: pull_request` and expanded the path filter. (`pull_request` also covers fork PRs and associates the run with the PR as a gate.)
- **`build.yml`**'s `build-cli` job published the generated CLI to `phrase-cli` **without ever compiling it**. Added a `go build` (against the freshly generated go client, so it doesn't depend on a possibly-lagging released `phrase-go`) before the publish step.

## Verification

- Regenerated the CLI: `screenshot/create` (required `filename`) and `upload/create` (required `file`) both emit correct `os.Open(...)`; no malformed output anywhere.
- Built `clients/cli` against the freshly generated `clients/go` — clean build.

## Note

The follow-on `too many arguments` error is a separate matter: the CLI's committed `go.mod` pins a released `phrase-go` that predates this endpoint. It resolves once a new `phrase-go` is released from the updated spec and the CLI bumps its dependency — unaffected by this PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -59,7 +59,12 @@ jobs:
           GOPRIVATE: github.com/phrase/phrase-go
         run: |
           npm install
+          npm run generate.go
           npm run generate.cli
+          # Verify the generated CLI compiles before publishing it. Build against
+          # the freshly generated go client so this does not depend on a released
+          # phrase-go version that may lag behind the spec.
+          (cd ./clients/cli && go mod edit -replace github.com/phrase/phrase-go/v4=../go && go build . && go mod edit -dropreplace github.com/phrase/phrase-go/v4)
           git clone https://$API_TOKEN_GITHUB@github.com/phrase/phrase-cli.git clones/cli &> /dev/null
           rsync -avI --delete --exclude='.git/' clients/cli/ clones/cli
           cd clones/cli
--- a/.github/workflows/test-cli.yml
+++ b/.github/workflows/test-cli.yml
@@ -5,7 +5,9 @@ on:
   pull_request:
     paths:
       - 'paths/**'
+      - 'paths.yaml'
       - 'schemas/**'
+      - 'schemas.yaml'
       - 'parameters.yaml'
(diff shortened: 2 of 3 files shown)


jablan:
Good catch — fixed in a632435. Both file branches (required and optional) now read with `helpers.ToSnakeCase("{{vendorExtensions.x-export-param-name}}")`, the same key `AddFlag` registers with, instead of `paramName`. The Go local variable name stays `paramName`; only the viper lookup key changed.

This was a pre-existing inconsistency in the optional `file` branch that I copied into the new required branch — it happened to work because `paramName == x-export-param-name` for `file`/`filename`, but would have silently dropped the upload for any escaped/disambiguated name. Regenerated and verified `screenshot/create` and `upload/create` still build clean.

sbl (approved):

forelabs (approved):

PR #1210, opened 2026-06-25
feat(API): Add translation_key_prefix and filter_by_prefix to async locale download

forelabs:
## Changes

- Add `translation_key_prefix` and `filter_by_prefix` params to `locale_downloads/create` for parity with the sync locale download endpoint
- Clarify `filter_by_prefix` description on both endpoints: it requires `translation_key_prefix` to be set

### DIFF
--- a/paths/locale_downloads/create.yaml
+++ b/paths/locale_downloads/create.yaml
@@ -107,6 +107,13 @@ requestBody:
               Custom metadata filters. Provide the name of the metadata field and the value to filter by. Only keys with matching metadata will be included in the download.
             type: object
             properties: {}
+          translation_key_prefix:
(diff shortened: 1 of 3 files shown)


paths/locale_downloads/create.yaml
```diff
@@ -107,6 +107,13 @@ requestBody:
               Custom metadata filters. Provide the name of the metadata field and the value to filter by. Only keys with matching metadata will be included in the download.
             type: object
             properties: {}
+          translation_key_prefix:
```
jablan, line 110: we should probably extract these to a reference as they are repeated in multiple endpoints. I'll probably try to do it in a separate PR

jablan (approved):

PR #1219, opened 2026-06-29
fix(CLI): publish binaries even when sign and notarize fails

jablan:
## Summary

- Moves the GitHub Release publish step into the `release` job so all binaries (linux, windows, macos) are always published, regardless of signing outcome
- Adds `continue-on-error: true` to `sign_and_notarize` — it still runs and overwrites the macOS assets with signed versions on success, but its failure no longer blocks the release
- Updates `brew` to `needs: [release, sign_and_notarize]` with `if: always() && needs.release.result == 'success'` so Homebrew bump waits for signing when possible but isn't blocked by signing failures

Fixes: https://github.com/phrase/phrase-cli/actions/runs/28353873876/job/83992891160

## Test plan

- [ ] Merge triggers a CLI release via release-please
- [ ] Verify binaries appear in GitHub Release even if sign_and_notarize were to fail


theSoenke (approved):

jablan:
closing as not sure if useful or not

PR #1223, opened 2026-07-23
fix(API): extract cldr_version enum to fix TypeScript client build

francawinter:
## What changed

Removed the `enum` constraint from `cldr_version` in the `POST /icu/skeleton` request body.

## Why

The `typescript-fetch` generator has a bug: inline string enums inside schemas that also use `oneOf` emit a property type reference without the matching enum declaration, causing a TypeScript compile error in the generated client:

```
src/models/IcuSkeletonParameters.ts: Cannot find name 'IcuSkeletonParametersCldrVersionEnum'
```

Removing the enum fixes the build. The description and example are updated to reflect the actual accepted values (`legacy` or a CLDR version string such as `cldr48`).

## Steps to verify

CI `jest` / Run Typescript Tests should pass.

### DIFF
--- a/schemas/icu.yaml
+++ b/schemas/icu.yaml
@@ -13,10 +13,7 @@ skeleton:
 skeleton_cldr_version:
   type: string
   description: >-
-    Pluralization rule set to apply when constructing skeletons.
-    Accepted values are `legacy` and `cldr_41`. Defaults to
-    `legacy` when omitted.
-  enum:
-    - legacy
-    - cldr_41
-  example: cldr_41
+    Pluralization rule set to use when constructing skeletons. Pass `legacy`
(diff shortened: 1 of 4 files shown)


schemas/icu.yaml
```diff
@@ -9,3 +9,14 @@ skeleton:
   example:
     en: "{number, plural, one {} other {}}"
     de: "{number, plural, one {} other {}}"
+
+skeleton_cldr_version:
+  type: string
+  description: >-
+    Pluralization rule set to apply when constructing skeletons.
+    Accepted values are `legacy` and `cldr_41`. Defaults to
```
theSoenke: not sure whether that's correct that only cldr 41 and legacy are accepted

theSoenke (approved):

PR #1242, opened 2026-08-26
feat(API): add Checks API endpoints

docstun:
## What

Adds REST API endpoints for the new Checks API:

- `GET /projects/{project_id}/checks/issues` — list check issues for a project, filterable by:
  - `state` — `active` (default), `solved`, `dismissed`, or `all`
  - `locale_ids[]`
  - `check_names[]`
  - standard `page` / `per_page` pagination
- `PATCH /projects/{project_id}/checks/issues/{id}/dismiss` — mark a check issue as dismissed so it drops off the default active list.

## Details

- New `check_issue` schema mirrors the API serializer: `id`, `check_name`, `state`, `description`, `dismissed_at`, `solved_at`, `created_at`, `updated_at`, and the embedded `translation`.
- New `Checks` tag added under the **Quality** tag group.
- Listing requires the `read` scope; dismissing requires the `write` scope.

Validated with `make lint` (swagger-cli valid, no openapi-generator validation errors).


paths/checks/dismiss.yaml
```diff
@@ -0,0 +1,57 @@
...
+        "$ref": "../../headers.yaml#/X-Rate-Limit-Reset"
+  '400':
+    "$ref": "../../responses.yaml#/400"
+  '404':
+    "$ref": "../../responses.yaml#/404"
+  '401':
+    "$ref": "../../responses.yaml#/401"
+  '403':
+    "$ref": "../../responses.yaml#/403"
+    description: Forbidden. Returned when the access token lacks the `write` scope or when the requesting user is not allowed to dismiss check issues in this project.
```
jablan: overriding references doesn't work. either it should be only a reference, or a fully defined node. :(

schemas/check_issue.yaml
```diff
@@ -0,0 +1,69 @@
...
+  example:
+    id: abcd1234cdef1234abcd1234cdef1234
+    check_name: translation_placeholder_usage
+    state: active
+    description: 'The translation contains invalid placeholders. (missing: %{count})'
+    dismissed_at: null
+    solved_at: null
+    created_at: '2015-01-28T09:52:53Z'
+    updated_at: '2015-01-28T09:52:53Z'
+    translation:
```
jablan, line 50: perhaps this could be a reference?

jablan (changes requested):
cosmetic issues pointed out

paths/checks/dismiss.yaml
docstun: Good catch — dropped the sibling `description` and kept the bare `$ref` to the shared 403 response. Fixed in 058d3e9.

schemas/check_issue.yaml
docstun: @jablan in examples, we so far always used the explicit notation, so I would not want to introduce this here.

jablan (approved):

PR #1258, opened 2026-09-08
docs(API): Document upload file_format and async state contract #STRINGS-3096

lookasc:
## Purpose

- Clarifies how the `POST /projects/{project_id}/uploads` endpoint's `file_format` parameter behaves, so API callers (including automated agents) stop guessing a format and silently failing.
- Documents that `file_format` is optional: when omitted, Phrase auto-detects the format from the file's content, which is the recommended approach.
- Notes that a provided `file_format` must match the uploaded file's actual content, not just the project's main format — a mismatch makes the upload fail.
- Documents the asynchronous contract: the endpoint returns `201` once the file is accepted, and callers should poll the upload's `state` field, where `error` signals a processing or parsing failure.
- Describes the possible upload `state` values so callers know what to expect while polling.

Documentation only — no schema types, required fields, or endpoint behavior change.

Related ticket: https://phrase.atlassian.net/browse/STRINGS-3096

### DIFF
--- a/paths/uploads/create.yaml
+++ b/paths/uploads/create.yaml
@@ -2,6 +2,8 @@ summary: Upload a new file
 description: |
   Upload a new language file. Creates necessary resources in your project.
 
+  The upload is processed asynchronously: this endpoint returns `201 Created` once the file has been accepted and enqueued, not once processing has finished. Poll `GET /projects/{project_id}/uploads/{id}` and check the `state` field — `error` means processing failed (for example, an unparseable file or a `file_format` that doesn't match the file's actual content).
+
   Note: be aware of [upload limits](https://support.phrase.com/hc/en-us/articles/8548271212188-Phrase-Strings-Limits#file-size-upload-limits-0-0).
 operationId: upload/create
 tags:
@@ -44,7 +46,6 @@ x-code-samples:
         -X POST \
         -F branch=my-feature-branch \
         -F file=@/path/to/my/file.json \
-        -F file_format=json \
         -F locale_id=abcd1234cdef1234abcd1234cdef1234 \
         -F tags=awesome-feature,needs-proofreading \
         -F locale_mapping[en]=2 \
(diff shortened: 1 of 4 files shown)


paths/uploads/create.yaml
```diff
@@ -83,7 +83,10 @@ requestBody:
             format: binary
             example: "/path/to/my/file.json"
           file_format:
-            description: File format. Auto-detected when possible and not specified.
+            description: |
+              File format of the uploaded file (e.g. `json`, `simple_json`, `yml`). Optional — when omitted, the format is auto-detected from the file's content. We recommend omitting this parameter and letting auto-detection handle it rather than guessing.
```
jablan: where does this recommendation come from? how do we detect e.g. different json formats?

lookasc: @jablan It is just a rewrite of what we had previously and my (incorrect) assumption based on what we use in the code `FormatGuesser.guess_format`.
I have just looked into its internals and know it is wrong for JSON formats.

What do you think to put it this way?

>               File format of the uploaded file, given as a format's `api_name`
>               When omitted, Phrase attempts o auto-detect the format from
>               the file's extension and content.
> 
>               Auto-detection is reliable only for formats with a distinctive
>               extension or structure (e.g. `yml`, `xliff`, `csv`, `xml`,
>               `strings`). It cannot disambiguate families that share an
>               extension — in particular the JSON formats (`json`, `simple_json`,
>               `nested_json`, `react_simple_json`, `go_i18n`, ...) all use `.json`
>               and differ only in structure. For those, set `file_format`
>               explicitly; otherwise the upload may be parsed as the wrong format
>               or rejected as ambiguous.
> 
>               A provided `file_format` must match the actual content of the
>               uploaded file, not merely the project's main format. A mismatch
>               makes the upload fail.

jablan: I don't think we should advise to omit this, on the contrary. I wouldn't clarify the auto detection beyond what's currently there. Also, the clarification about the project's main format seems too much, but obviously someone had problems with that. 🤷 
Perhaps we should also point to the list of supported formats: https://support.phrase.com/hc/en-us/articles/9652464547740-Supported-File-Formats-Strings, although that page doesn't list the format identifiers which should be used here.

jablan (approved):

